PR #8, opened 2019-04-09
[DOC] Added description of get_rapids_dataset_root_dir() and how both devel…

rlratzel:
…opers and end users can use it for specifying the location of datasets.

### DIFF
--- a/maintainers/testing.md
+++ b/maintainers/testing.md
@@ -89,25 +89,7 @@ Overall goal, follow the open source ecosystem for infrastructure choices
 
 ## Datasets
 
-- Many tests depend on the presence of specific datasets in order to properly verify code correctness.
-- Some datasets are small enough that they can be included with the source code, but others are relatively large and cannot practically coexist with the source. In both cases, the test executable should not assume datasets are located in a particular place (either with a relative path, since the executable could be moved, or an absolute path, since different machines will likely use different absolute paths), so the location of the datasets must be provided at test runtime.
-- There are three mechanisms that allow tests to locate datasets on disk:
-  - Users can set the environment variable `RAPIDS_DATASET_ROOT_DIR` to the path of the directory containing the datasets needed by tests.
-  - Users can pass the flag `-DRAPIDS_DATASET_ROOT_DIR=<root dir>` to the `cmake` command when building from source. If the environment variable described above is not set, this setting will be used.
-  - Finally, test developers can specify a default directory to use in the event that both the environment variable and the build option are not set. Since this is essentially hardcoding a path in the test executable, users should be made aware of what the location is as well as the options they have (the `RAPIDS_DATASET_ROOT_DIR` env var and the `-DRAPIDS_DATASET_ROOT_DIR=<root dir>` build option) for overriding it.
-
-### Test developer API
-
-- `const std::string& get_rapids_dataset_root_dir(const std::string& defaultRdrd)`
-  - Tests that need to access the dataset directory should use this function for properly retrieving the dataset root directory defined by the user at build-time or runtime, or a specific default hardcoded in the test passed in as `defaultRdrd`.
-  - An example is shown below:
-```
-    const std::string& rapidsDatasetRootDir = get_rapids_dataset_root_dir("/datasets");
-    readDataset(rapidsDatasetRootDir + "/golden_data/web-BerkStan.pagerank_val_0.85.bin");
-```
-In the above example, if the tests are built with `-DRAPIDS_DATASET_ROOT_DIR=/bar` and the environment variable `RAPIDS_DATASET_ROOT_DIR` is set to `/foo`, the `readDataset()` call will get `/foo/golden_data/web-BerkStan.pagerank_val_0.85.bin`. If the environment variable is not set, the call will get `/bar/golden_data/web-BerkStan.pagerank_val_0.85.bin`. Finally, if the user didn't set the environment var and didn't use the build flag, the call will get `/datasets/golden_data/web-BerkStan.pagerank_val_0.85.bin`.
-
-*NOTE: At the moment, only the `cugraph` library has adopted the `get_rapids_dataset_root_dir()` functionality.*
+- Many tests depend on the presence of specific datasets in order to properly verify code correctness. See [datasets]({{ site.url }}{{ site.baseurl }}/maintainers/datasets) for more details.
 
 ## Integration / Workflow Testing and Benchmarking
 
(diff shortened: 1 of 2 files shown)


maintainers/testing.md
```diff
@@ -87,6 +87,28 @@ Overall goal, follow the open source ecosystem for infrastructure choices
...
+- There are three mechanisms that allow tests to locate datasets on disk:
+  - Users can set the environment variable `RAPIDS_DATASET_ROOT_DIR` to the path of the directory containing the datasets needed by tests.
+  - Users can pass the flag `-DRAPIDS_DATASET_ROOT_DIR=<root dir>` to the `cmake` command when building from source. If the environment variable described above is not set, this setting will be used.
+  - Finally, test developers can specify a default directory to use in the event that both the environment variable and the build option are not set. Since this is essentially hardcoding a path in the test executable, users should be made aware of what the location is as well as the options they have (the `RAPIDS_DATASET_ROOT_DIR` env var and the `-DRAPIDS_DATASET_ROOT_DIR=<root dir>` build option) for overriding it.
```
harrism: Hmmm, shouldn't there be a default value for `RAPIDS_DATASET_ROOT_DIR` so that test developers don't *have* to configure it just in case it's not set?  Convention over configuration with option to override...

maintainers/testing.md
```diff
@@ -87,6 +87,28 @@ Overall goal, follow the open source ecosystem for infrastructure choices
...
+
+### Test developer API
+
+- `const std::string& get_rapids_dataset_root_dir(const std::string& defaultRdrd)`
```
harrism: Where is this API defined? This is only the docs repo.

rlratzel: https://github.com/rapidsai/cugraph/pull/211

maintainers/testing.md
rlratzel: That's probably a good idea. This replaces hardcoded instances of `/datasets`, so I suppose that's a decent default given the history. Thoughts?

Another option is to throw an exception if unset, forcing the user to know about it. This might not be too bad given that the build option also exists (a one time set vs. an env var).

maintainers/testing.md
```diff
@@ -87,6 +87,28 @@ Overall goal, follow the open source ecosystem for infrastructure choices
     - I.E. cuDF PyTests:
       - [https://github.com/rapidsai/cudf/tree/branch-0.7/python/cudf/tests](https://github.com/rapidsai/cudf/tree/branch-0.7/python/cudf/tests)
 
+## Datasets
+
+- Many tests depend on the presence of specific datasets in order to properly verify code correctness.
+- Some datasets are small enough that they can be included with the source code, but others are relatively large and cannot practically coexist with the source. In both cases, the test executable should not assume datasets are located in a particular place (either with a relative path, since the executable could be moved, or an absolute path, since different machines will likely use different absolute paths), so the location of the datasets must be provided at test runtime.
+- There are three mechanisms that allow tests to locate datasets on disk:
```
raydouglass: The descriptions of each mechanism are long. Why not state something like: "Tests can locate datasets on disk using three methods which are tried in order until one returns non-null" Then remove the lines in each about when it triggers.

maintainers/testing.md
```diff
@@ -87,6 +87,28 @@ Overall goal, follow the open source ecosystem for infrastructure choices
     - I.E. cuDF PyTests:
       - [https://github.com/rapidsai/cudf/tree/branch-0.7/python/cudf/tests](https://github.com/rapidsai/cudf/tree/branch-0.7/python/cudf/tests)
 
+## Datasets
+
+- Many tests depend on the presence of specific datasets in order to properly verify code correctness.
```
raydouglass: I don't think this whole section should be a bulleted list. Maybe take the first two bullets, condense them down and put under `## Datasets` as an overview sentence or two.

maintainers/testing.md
```diff
@@ -87,6 +87,28 @@ Overall goal, follow the open source ecosystem for infrastructure choices
...
+- There are three mechanisms that allow tests to locate datasets on disk:
+  - Users can set the environment variable `RAPIDS_DATASET_ROOT_DIR` to the path of the directory containing the datasets needed by tests.
+  - Users can pass the flag `-DRAPIDS_DATASET_ROOT_DIR=<root dir>` to the `cmake` command when building from source. If the environment variable described above is not set, this setting will be used.
+  - Finally, test developers can specify a default directory to use in the event that both the environment variable and the build option are not set. Since this is essentially hardcoding a path in the test executable, users should be made aware of what the location is as well as the options they have (the `RAPIDS_DATASET_ROOT_DIR` env var and the `-DRAPIDS_DATASET_ROOT_DIR=<root dir>` build option) for overriding it.
+
+### Test developer API
+
+- `const std::string& get_rapids_dataset_root_dir(const std::string& defaultRdrd)`
+  - Tests that need to access the dataset directory should use this function for properly retrieving the dataset root directory defined by the user at build-time or runtime, or a specific default hardcoded in the test passed in as `defaultRdrd`.
```
raydouglass: "Tests should use this function which return the dataset root directory based on the rules above."

maintainers/testing.md
```diff
@@ -87,6 +87,28 @@ Overall goal, follow the open source ecosystem for infrastructure choices
...
+  - An example is shown below:
```
raydouglass: Maybe make this a subheading: `#### Example API usage`

mike-wendt:
I'd like to move this to it's own MD file, as it will get lost in that larger file. @rlratzel can you break it into a separate file? I'll work on the tree and everything

maintainers/testing.md
```diff
@@ -87,6 +87,28 @@ Overall goal, follow the open source ecosystem for infrastructure choices
...
+```
+    const std::string& rapidsDatasetRootDir = get_rapids_dataset_root_dir("/datasets");
+    readDataset(rapidsDatasetRootDir + "/golden_data/web-BerkStan.pagerank_val_0.85.bin");
+```
+In the above example, if the tests are built with `-DRAPIDS_DATASET_ROOT_DIR=/bar` and the environment variable `RAPIDS_DATASET_ROOT_DIR` is set to `/foo`, the `readDataset()` call will get `/foo/golden_data/web-BerkStan.pagerank_val_0.85.bin`. If the environment variable is not set, the call will get `/bar/golden_data/web-BerkStan.pagerank_val_0.85.bin`. Finally, if the user didn't set the environment var and didn't use the build flag, the call will get `/datasets/golden_data/web-BerkStan.pagerank_val_0.85.bin`.
```
raydouglass: I think you can make this more clear if it was a table. Columns could be `-DRAPIDS_DATASET_ROOT_DIR`, `RAPIDS_DATASET_ROOT_DIR`, and the result.

maintainers/testing.md
rlratzel: Agreed. I incorporated your suggestion in the separate document that Mike mentioned I should create.

maintainers/testing.md
rlratzel: I rewrote how the mechanisms are used in the new document and I think it's clearer now.

maintainers/testing.md
rlratzel: The latest API uses a default value now of `/datasets`, which users can override with either the build option or the env var, or both.  https://github.com/rapidsai/cugraph/pull/215

maintainers/testing.md
rlratzel: I rewrote how the mechanisms are used in the new document and I think it's clearer now.

maintainers/testing.md
rlratzel: Agreed, done.

maintainers/testing.md
rlratzel: I rewrote how the mechanisms are used in the new document and I think it's clearer now.

mike-wendt:
I'll do a full review, tonight. I'm debating on if this should be in a subsection or not and I just need to pull your code to try it out. I won't be able to do it until later so I'll let you know what I decide then.

PR #23, opened 2019-05-23
Updated get started links

exactlyallan:
Minor text and link changes to get started docs page.
@mike-wendt will 'latest' be removed as a tag from our github repos? We should make note on where on the site and docs we would need to update link for new releases.


exactlyallan:
@mike-wendt Can you merge plz?

PR #37, opened 2019-08-08
Add burn down guide

raydouglass:
Add some documentation on how to perform the start of burn down from the Operation perspective.


fondaing:
looks great

PR #47, opened 2019-10-10
[REVIEW] Update gpuCI docs with skipci caveat

dillon-cullinan:


pentschev:
Nice, thanks for documenting this @dillon-cullinan !

PR #74, opened 2020-03-18
[RELEASE] New Docs Site and Release 0.13 Updates

mike-wendt:
Staging for new docs site changes addressed in #71 and updates for v0.13


mike-wendt:
✅ approved (given that we're unable to do it officially due to the size)

PR #76, opened 2020-03-19
Link assets instead of inlining them

ajschmidt8:
This PR includes the following changes:

- Changes the customization script to include a link to remote CSS/JS assets instead of inlining them
   - Updates docs to reflect this change.
- Move contents of `rapids.js` and `rapids.css` into `assets/` folder and delete original files
- Minor css tweaks to adjust scoping
- Makes all paths relative
- Removes `<script>` tags that reference non-existent files


mike-wendt:
@ajschmidt8 seems like the review is happening in #77 for this. Do we need to keep this PR?

ajschmidt8:
> @ajschmidt8 seems like the review is happening in #77 for this. Do we need to keep this PR?

nope, this can be closed. just opened this since the other PR can't easily be reviewed.

PR #96, opened 2020-06-19
[REVIEW] Update conda command flag

dillon-cullinan:
`-f` option doesn't exist, replacing in docs with `--force` which is the correct option.


rlratzel:
This is a good change, not only because it's the right flag, but using long options in scripts and documentation in general is preferred for making them self-documenting. We should probably have that policy across the board IMO.

PR #97, opened 2020-06-23
Remove nvstrings

ajschmidt8:
This PR removes `nvstrings` and `libnvstrings` from the nightly docs since `nvstrings` is now a part of `cudf`.

Specifically, the following changes have been made:

- Set the `nightly` parameter to `0` for `nvstrings` and `libnvstrings` in `_data/docs.yml`
- Remove the following folders:
  - `api/libnvstrings/nightly `
  - `api/libnvstrings/0.15 `
  - `api/nvstrings/nightly `
  - `api/nvstrings/0.15`


ajschmidt8:
looks like there's an issue with the links. fixing now

ajschmidt8:
fixed in https://github.com/rapidsai/docs/pull/97/commits/419c9663a9f45d1860a3a823e16132dbaca0fbcc.

There's a hack in that commit, so I'll explain it here so it's archived. The hack was necessary mostly because Jekyll's [Liquid](https://shopify.github.io/liquid/) templating engine isn't super capable. The hack also required me to change `_data/releases.yml` to a dictionary, but that change seemed fine given the variable names/relationship. I updated all the necessary templates to reflect this change.

---

Given `api.versions` is a dictionary like this:

```yaml
legacy: 1
stable: 1
nightly: 0
```

This hack: 
```
{% assign versions = api.versions | sort | where_exp: "item", "item[1] == 1" | join: "" | split: "1" | reverse %}
```

converts it in the following manner:

1. `sort` converts `{"legacy": 1},{"stable": 1},{"nightly": 0}` to `[["legacy", 1],["nightly", 0],["stable", 1]]`
2. `where_exp` filters out any key/value pairs that aren't 1, so it returns `[["legacy", 1],["stable", 1]]`
3. `join` returns the string `legacy1stable1`
4. `split` returns `["legacy", "stable"]`
5. `reverse` returns `["stable", "legacy"]`.

PR #103, opened 2020-07-20
ENH Add RAPIDS Notices to docs site

mike-wendt:
The RAPIDS Notices will be used to communicate and document important changes to the project. There are [three types](https://github.com/rapidsai/docs/blob/enh-rapids-notices/notices/notices.md#notice-types) of notices to target different users and impact areas. Included in this PR are 3 urgent notices and 1 historical notice (RSN 1).

PR preview link for changes: https://deploy-preview-103--docs-rapids-ai.netlify.app/notices

In addition the `_config.yaml` was updated to fix the issues with local dev and the `api/` folder. This moved the `api.md` folder to the root so all of the `api/` folder can be ignored. Now we doing development locally the re-render times are back to a few seconds instead of 30+ seconds.


mike-wendt:
Still need to figure out the duplicate "API Docs" on the left nav

mike-wendt:
> Still need to figure out the duplicate "API Docs" on the left nav

Finally fixed this

_config.yml
```diff
@@ -8,9 +8,21 @@ baseurl: "/" # the subpath of your site, e.g. /blog
...
+  - update_symlinks.sh
+  - _config_ignore_api.yml
+include:
+  - _sources
+  - _static
+  - _images
+
+collections:
+  notices:
+    output: ture
```
ajschmidt8, line 25: should this be `true`?

_config_ignore_api.yml
```diff
@@ -0,0 +1,57 @@
...
+  - _config_ignore_api.yml
+  - api/
+include:
+  - _sources
+  - _static
+  - _images
+
+collections:
+  notices:
+    output: ture
```
ajschmidt8, line 26: should this also be `true`?

PR #108, opened 2020-08-04
[REVIEW] Docs for git lfs

raydouglass:


resources/git.md
```diff
@@ -107,3 +107,17 @@ DOC Update conda install process
...
+
+Any file larger than 5MB must be stored using either [Git LFS](https://git-lfs.github.com/) or S3.
+
+| Size | Required | Not Required |
+| ---- | -------- | ------------ |
+| <5MB | Git | Git |
+| >5MB | Git LFS | S3 |
+| >2GB | Avoid | S3 |
+
+A required file is a file that the average developer requires in order to build and/or test the project. This may include small datasets for running unit tests or large source files.
```
mike-wendt, line 121: I would add quotes or a tick mark around the "required" work to make it stand out. It seems almost clear enough but we may want to be more explicit with some 80/20 rule or something to make it clear that larger files that aren't needed by the majority should be in S3 for the simple ability to "opt-in" to download those. In doing so I think we can highlight the fact that "Git LFS" has consequences when it is used as a "catch-all" and impacts devs in the long term.

I would also add that we need to make the right decisions on these files before they are merged. Once merged they cannot be switched to LFS without impacting the git history.

PR #125, opened 2020-09-15
DOC Add RGN 5 for RAPIDS/Colab support

mike-wendt:
Colab support only supports Python 3.6 which was dropped in RAPIDS `0.15`. This notice is classified as `RGN` as it is a breaking change and there are existing `RSN` notices that are linked in this article.

Preview of notice RGN 5 - https://deploy-preview-125--docs-rapids-ai.netlify.app/notices/rgn0005/


raydouglass (approved):

aravenel:
@datametrician You OK with this language?

datametrician (approved):

datametrician:
> @datametrician You OK with this language?

Yep. It's cut and dry and factual.

PR #132, opened 2020-10-09
Add 404 Page and Remove Old Docs

ajschmidt8:
This PR removes all of the docs that are older than the current legacy version, `0.14`. Additionally, it adds a `404.md` page that includes some `js` to redirect users to the `/api` page if they 404 on a `/api` sub-page. Finally, it adds `.jekyll-cache/` to the `.gitignore` file.

**Note:** This PR removes thousands of files and is therefore very large. It's probably easiest to view it by commit rather than clicking on the `Files Changed' tab.


ajschmidt8:
keeping this as a draft to prevent accidental merges before it's reviewed.

mike-wendt (approved):

PR #144, opened 2020-12-08
[REVIEW] ENH Move RMM to librmm folder

dillon-cullinan:
This PR simply moves all the C++ RMM docs to `librmm` to make space for publishing RMM python docs.


ajschmidt8 (changes requested):
Can you update the section below too? You can use the `libcuml` code block above it as an example. Basically just need to change the key value, `name`, and `path` to `librmm`.

https://github.com/rapidsai/docs/blob/gh-pages/_data/docs.yml#L130-L140

ajschmidt8 (approved):

PR #169, opened 2021-02-26
FIX Correct error from docs release process

mike-wendt:
The docs release process should have set legacy to 0.17 but it remained 0.16 after removing the 0.16 files. This caused Jekyll and the GH pages builds to fail on deployment


mike-wendt:
@ajschmidt8 FYI not sure why this happened, can you investigate when you have time?

PR #221, opened 2021-08-12
Include list and struct into basic.html

Nicholas-7:
Incorporating List and Struct to the basics.html file


ajschmidt8:
@Nicholas-7, the `html` files in this repo are automatically generated from the RAPIDS libraries themselves. In order for these changes to persist, you will need to add them to the `cudf` repo itself. Otherwise they will just get overwritten the next time we generate the docs (which happens on nightly basis).

Because of this, I will close this PR.

PR #247, opened 2022-01-27
Add CSS rule to fix PyData theme

ajschmidt8:
This PR fixes the CSS issue reported in https://github.com/rapidsai/cudf/issues/9761. The sidebar height needs to be updated (increased) to account for the additional HTML elements that we add to the page as part of our post-processing.

### DIFF
--- a/assets/css/custom.css
+++ b/assets/css/custom.css
@@ -497,3 +497,15 @@ a.rapids-selector__menu-item--selected {
   vertical-align: inherit !important;
   font-weight: bold;
 }
+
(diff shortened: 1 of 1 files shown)


ajschmidt8:
FYI, the changes in this PR are just an override for the default sidebar height (for our version of PyData) that's listed below:

https://github.com/pydata/pydata-sphinx-theme/blob/3532922d15654ce82f84c8e1360720c8713af524/src/scss/index.scss#L226-L234

Ethyling (approved):

PR #269, opened 2022-05-19
Update Dask + Pandas to Dask + cuDF path

miguelusque:
According to https://github.com/rapidsai/cudf/issues/3176#issuecomment-545582810, the documentation is outdated. I have amended it.


ajschmidt8:
@miguelusque, this documentation is generated from `cudf`'s source code. It will need to be edited in the `cudf` repository in order for this change to persist. Otherwise it will just be overwritten the next time we build all the docs. I will close this PR.

PR #272, opened 2022-06-02
Add `libcuspatial` to docs page

ajschmidt8:
This PR adds `libcuspatial` to the docs index page since they've recently started publishing documentation.

### DIFF
--- a/_data/docs.yml
+++ b/_data/docs.yml
@@ -172,6 +172,17 @@ libs:
       legacy: 1
       stable: 1
       nightly: 1
+  libcuspatial:
+    name: libcuspatial
+    path: libcuspatial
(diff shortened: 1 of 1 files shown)


_data/docs.yml
```diff
@@ -172,6 +172,17 @@ libs:
       legacy: 1
       stable: 1
       nightly: 1
+  libcuspatial:
+    name: libcuspatial
+    path: libcuspatial
+    desc: 'libcuspatial is a GPU accelerated C/C++ library for accelerating GIS workflows including point-in-polygon, spatial join, coordinate systems, shape primitives, distances, and trajectory analysis.'
```
isVoid, line 178: Should we keep the description the same as: https://github.com/rapidsai/cuspatial/blob/branch-22.08/cpp/doxygen/main_page.md ?

ajschmidt8: sure! Didn't see that. updated in dc61a5a

PR #281, opened 2022-07-15
add RSN re: pinning cuda python

taureandyernv:
Create an RSN to address the cuda python issue and possible resolution actions

### DIFF
--- a/_notices/rsn0020.md
+++ b/_notices/rsn0020.md
@@ -1,0 +1,63 @@
+---
+layout: notice
+parent: RAPIDS Support Notices
+grand_parent: RAPIDS Notices
+nav_exclude: true
+notice_type: rsn
+# Update meta-data for notice
+notice_id: 20 # should match notice number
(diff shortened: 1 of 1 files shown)


jarmak-nv:
Super minor visualization comment, in the `Status` section I think it might read a little easier if we switch it to:

## Status

Currently, we have these two resolution paths to offer:

#### Resolution 1: If you can upgrade RAPIDS
- Please upgrade your RAPIDS version to `v22.06.01` using our [Release Selector](https://rapids.ai/start.html#get-rapids)

#### Resolution 2: If you can NOT upgrade RAPIDS:
- Pin `cuda-python` to `11.7.0` by using `cuda-python=11.7.0` when installing  

- Example command: `conda install -c nvidia cuda-python=11.7.0`.

Older versions of RAPIDS may also be affected and should attempt to upgrade or pin.

aravenel (approved):

PR #282, opened 2022-07-19
Make RSN 0020 clearer

taureandyernv:
Added some language updates to make it clearer on what is pinned and what is the problem in a few places


_notices/rsn0020.md
```diff
@@ -64,4 +64,5 @@ Older versions of RAPIDS may also be affected and users should attempt to upgrad
 ## Impact
 
 This issue impacts users trying to install RAPIDS or install additional libraries that may 
-require an update to `cuda-python` and select `11.7.1`.
\ No newline at end of file
+require an update to `cuda-python` and select `11.7.1`, as version `11.7.1` will call one 
+of its dependencies that breaks RAPIDS.
```
bdice: Do you think this would be more helpful?

> This issue impacts users installing conda packages of RAPIDS 22.02 through 22.06, which depend on `cuda-python`. This issue also affects existing conda installations of RAPIDS versions 22.02 - 22.06 if the user installs additional packages depending on `cuda-python` that result in an update to `cuda-python>11.7.0`.

aravenel (approved):

PR #300, opened 2022-10-05
Create Notice Cuda 11 EOL 

caryr35:
This PR creates CUDA 11 EOL

### DIFF
--- a/_notices/rsn0017.md
+++ b/_notices/rsn0017.md
@@ -6,7 +6,7 @@ nav_exclude: true
 notice_type: rsn
 # Update meta-data for notice
 notice_id: 17 # should match notice number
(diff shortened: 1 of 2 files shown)


_notices/rsn0017.md
```diff
@@ -9,8 +9,8 @@ notice_id: 17 # should match notice number
 notice_pin: true # set to true to pin to notice page
```
aravenel, line 9: Should we unpin this one and pin the following? We have a lot of RSNs pinned, not sure we need them all.

raydouglass (approved):

_notices/rsn0017.md
raydouglass: I agree with this

PR #341, opened 2023-02-24
Adding a note on draft PRs

mmccarty:
Following up after the 23.02 retro to add a note to the maintainers docs on draft PRs.

PR #319, opened 2022-12-15
Update to Docs Home to sync with Rapids.ai Update [WIP]

exactlyallan:
Removing outdated sections and information, adding new sections 'install' and 'user-guide' to sync with Rapids.ai refactor https://github.com/rapidsai/rapids.ai/pull/271

PR #306, opened 2022-11-04
Update PRs documentation.

bdice:
Resolves https://github.com/rapidsai/ops/issues/2410.

PR #344, opened 2023-03-06
Restored contributing docs

mmccarty:
- [x] Restored contributing docs from #306 that were removed in #319
- [x] Added note on draft PRs from #341 
- [x] Added link in forward-merger docs.

### DIFF
--- a/contributing/code.md
+++ b/contributing/code.md
@@ -1,0 +1,78 @@
+---
+layout: default
+nav_order: 2
+parent: Contributing
+title: Code Contributions
+---
+
+# Code Contributions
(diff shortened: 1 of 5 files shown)


exactlyallan (approved):

bdice (approved):
Thanks @mmccarty. We should merge this to re-add the contributing guide, but I think we'll want to rework some outdated pieces of the content. That would be good to do at the same time as a revamp for each repository's CONTRIBUTING.md, so that we can consolidate some content and establish a welcoming and consistent guide for contributors.

PR #347, opened 2023-03-13
correct pip and colab template links

taureandyernv:
quick correction on the links as they were switched on the site


exactlyallan (approved):
My bad, thanks.

PR #362, opened 2023-04-14
Add S3 workflow

ajschmidt8:
This PR adds a workflow to deploy our site to production.

The workflow works by:

- Building the Jekyll site
- Downloading the library documentation files from S3 and injecting them into the Jekyll build directory
- Post-processing the S3 docs
- Deploying the site to Netlify

Additionally, it includes some additional changes to accommodate this new repository workflow. Specifically:

- adds a `.ruby-version` file to the repository
  - this file is used to ensure that the GHAs deploy workflow and the Netlify build preview environments use the same ruby version
- removes outdated references to files that used to be committed to the `api/` folder
- updates the `release_checklist.md` file to remove info about post-processing since this is handled by GHAs now
- removes any `link` directives in Jekyll that point to `api/` files since they're no longer committed to the repository and therefore cause build errors during `jekyll build`

**Note:** One of the consequences of this change is that `api/` files will not be available in Netlify deploy previews on pull requests.


ajschmidt8:
I'd like to benchmark this workflow with self-hosted runners, but apparently `ruby/setup-ruby` problems with self-hosted runners:

- https://github.com/ruby/setup-ruby#using-self-hosted-runners

I'll save that improvement for another day.

raydouglass (approved):

PR #357, opened 2023-04-11
Update GHAs self-hosted runner docs for `arc-v2`

ajschmidt8:
## Changes

This PR updates our GitHub Action self-hosted runner docs in anticipation of our [actions-runner-controller](https://github.com/actions/actions-runner-controller/) `v2` rollout.

## Background Context

One of the big changes from `v1` to `v2` is related to how self-hosted runner labels are handled.

`v1` allowed us to assign multiple labels to a particular runner set. For example `gpu-a100-525-1` and `gpu-a100-latest-1` could point to the same set of runners.

For `v2`, this is not possible. From speaking with the `actions-runner-controller` devs, this change is intentional. There is an open issue about it [here](https://github.com/actions/actions-runner-controller/issues/2445) that I am tracking.

As a result of this change, we had to decide whether to strictly use either `gpu-a100-525-1`-type labels or `gpu-a100-{latest,earliest}-1`-type labels.

Because of our finite GPU capacity and the overhead associated with manually updating driver-specific labels (e.g. `gpu-a100-525-1`) across all supported GitHub organizations, we've opted to adopt using the `earliest`/`latest` label types.

We will update our runners accordingly if this changes in the future.

PR #367, opened 2023-04-19
Update GHAs self-hosted runner docs for `arc-v2`

ajschmidt8:
Duplicate of #357 now that the force push is complete.


ajschmidt8:
(Duplicate of #357 now that the force push is complete.)

PR #372, opened 2023-04-21
Pip install corrections

jarmak-nv:
## Overview
Currently, there are two `#pip` anchors. This PR fixes the duplicate anchor issue in `install.md` found by @bdice 

## Discussion
This has been cut for future work in https://github.com/rapidsai/rapids.ai/issues/306
~However, I would like to discuss the potential removal of the content within Section 2a -> pip~. 

Currently all it does is have some text saying:
> The package installer for python (pip) is available as of RAPIDS release 23.04. See below for details.

IMO it's a better experience to remove that text and have the link in section 2 go straight to the pip install section.

### DIFF
--- a/install/install.md
+++ b/install/install.md
@@ -17,7 +17,7 @@ RAPIDS has several methods for installation, depending on the preferred environm
 **[2A. Setup Environment](#setup)** <br/>
 **[2B. Setup WSL2 Environment](#WSL2)** <br/>
 **[3A. Install RAPIDS (Release Selector)](#selector)** <br/>
-**[3B. Install RAPIDS (pip)](#pip)** <br/>
+**[3B. Install RAPIDS (pip)](#pip-install)** <br/>
 **[4. Getting Started](#start)**  
 
 <hr/>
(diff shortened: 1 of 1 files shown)


bdice:
@jarmak-nv I'm in favor of removing the environment section here. It seemed pretty useless to me too, but I couldn't decide if that deserved more thought for the overall flow of this document.

install/install.md
```diff
@@ -17,7 +17,7 @@ RAPIDS has several methods for installation, depending on the preferred environm
 **[2A. Setup Environment](#setup)** <br/>
 **[2B. Setup WSL2 Environment](#WSL2)** <br/>
 **[3A. Install RAPIDS (Release Selector)](#selector)** <br/>
-**[3B. Install RAPIDS (pip)](#pip)** <br/>
+**[3B. Install RAPIDS (pip)](#pip-install)** <br/>
```
bdice, line 20: If pip is now part of the release selector, why do we need "3A Selector" and "3B pip"?

bdice: Maybe the pip prerequisites should be in Section 2A "pip" instead of Section 3B?

bdice: Also "setup" is two words if you don't mind adjusting that... https://writingexplained.org/setup-vs-set-up-difference

bdice: If I can have one more request on this comment thread... can we use a nested structure for this "table of contents"? Something that looks more like:

image

bdice: @jarmak-nv We can defer some or all of these ideas to a subsequent PR. I'm happy for this PR to just fix the link.

jarmak-nv: Totally agree with these changes, but let's push them over to align with the work in https://github.com/rapidsai/rapids.ai/issues/306 

I'll collect these comments and put them into that issue.

I've updated the setup wording since it's a no-discussion required fix, but the rest will take a bit of back and forth I imagine.

jarmak-nv: Moved these comments over. I think we can resolve this conversation, but leave it to you @bdice if you have followups.

bdice: Sweet! I appreciate that. I had seen that issue but forgot about it.

bdice (approved):
LGTM. Thank you for doing this, and thank you for separating "easy" fixes in this PR from the broader discussion in https://github.com/rapidsai/rapids.ai/issues/306.

ajschmidt8 (approved):

PR #383, opened 2023-05-08
Update WSL pip docs.

### LINKED ISSUE #380: [DOC] pip requirements are unclear
## Report incorrect documentation

**Location of incorrect documentation**
https://docs.rapids.ai/install#pip-install

**Describe the problems or issues found in the documentation**

I want to do some revisions of the pip requirements, which I have helped several users with recently. See: https://github.com/rapidsai/cudf/issues/13233#issuecomment-1524535099 for an example and more details. We can do this in this PR #377 or a follow-up PR. For the moment, I am going to list two clarifications that are needed:

- Users need a **CUDA 11 toolkit** to use pip packages. CUDA 12 will fail.
  - We need to fix our docs where it says "CUDA >= 11.2, with at least the v460.27.03 driver" because that implies CUDA 12 is valid. It is not, for pip.
- We need to increase emphasis (even more) on the WSL side, to **ensure that CUDA drivers are not installed in WSL**; only `cuda-toolkit` and not `cuda`. We already point to a guide that explains this and we say in the link text, "[install the CUDA Toolkit without drivers](https://developer.nvidia.com/cuda-11-8-0-download-archive?target_os=Linux&target_arch=x86_64&Distribution=WSL-Ubuntu&target_version=2.0&target_type=deb_local) into the WSL2 instance."
  - I think we can spell out exactly what this means in the requirements step: we should say "Execute `sudo apt-get -y install cuda-toolkit` and not `sudo apt-get -y install cuda` to avoid installing a driver into WSL2."

bdice:
Resolves #380.

### DIFF
--- a/install/install.md
+++ b/install/install.md
@@ -10,7 +10,7 @@ description: |
 # RAPIDS Installation Guide
 {: .fs-8 }
 
-RAPIDS has several methods for installation, depending on the preferred environment and version. <br/>
+RAPIDS has several methods for installation, depending on the preferred environment and version.
 *New Users* should review the system and environment prerequisites.
 
 **[Install RAPIDS with Release Selector](#selector)** <br/>
--- a/install/install.md
+++ b/install/install.md
@@ -11,8 +11,7 @@ description: |
 {: .fs-8 }
 
 RAPIDS has several methods for installation, depending on the preferred environment and version. <br/>
-*New Users* should be mindful of the system and environment prerequisites. 
-{: .fs-6 .fw-300 }
+*New Users* should review the system and environment prerequisites.
 
 **[Install RAPIDS with Release Selector](#selector)** <br/>
 - [Installation Troubleshooting](#troubleshooting)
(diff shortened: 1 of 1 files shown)


install/install.md
```diff
@@ -11,8 +11,7 @@ description: |
 {: .fs-8 }
 
 RAPIDS has several methods for installation, depending on the preferred environment and version. <br/>
-*New Users* should be mindful of the system and environment prerequisites. 
-{: .fs-6 .fw-300 }
```
bdice, line 15: The phrase "be mindful of" sounds a little bit intense to me. Like "be mindful of _&lt;something dangerous&gt;_."

Also the font size for this felt aggressively large. It's text, not a heading.

install/install.md
```diff
@@ -11,8 +11,7 @@ description: |
 {: .fs-8 }
 
 RAPIDS has several methods for installation, depending on the preferred environment and version. <br/>
```
bdice, line 13:
```suggestion
RAPIDS has several methods for installation, depending on the preferred environment and version.
```

ajschmidt8 (approved):

PR #389, opened 2023-05-19
Add CLX deprecation notice to user guide and api doc pages

efajardo-nv:
CLX was deprecated in 23.04. Update user guide and and api doc pages to indicate this along with links to Morpheus.


bdice (approved):
Thanks @efajardo-nv! Looks good to me.

ajschmidt8 (approved):

Issue #386, opened 2023-05-16
Update docs for CUDA 12 pip wheel release


vyasr:
The core RAPIDS packages that build wheels (rmm, cudf, raft, cuml, cugraph, cuopt, ucx-py) are now also building wheels with CUDA 12 support. We plan to release these packages in 23.06. The documentation needs to be updated to reflect that release.

CC @exactlyallan @ajschmidt8

jarmak-nv:
This is done with #394 and #396 :tada:

PR #394, opened 2023-05-31
Release Selector 23.06 Updates

jarmak-nv:
This PR updates the release selector for 23.06:
- Python `3.8` has been replaced with `3.9`
- CUDA 12 has been added for pip!
   - In Conda/Docker it shows as a disabled button
   - I've added a new CUDA options row for pip - doing this let us get rid of the note and make it clear the CUDA version supported directly through the selector
- Moves `Method` above `Python` since it's more thematically appropriate IMO
   - ~I could see doing this for CUDA also, but didn't want to change _tooooo_ much all at once~
   - Made the change - I did this because `Method` changes the results of `CUDA`, feels wrong for something downstream to be above physically
- Adds cuSpatial for pip (Paul's working hard here 🙏 )
- Removes all `CLX` options as it has been removed/archived from RAPIDS

Some thoughts/comments:
- I don't clear the conda/pip cuda versions when they aren't being used - I figured it's a better experience for it to be saved if the user goes back and forth (and seemed like wasted cycles)
- I think we might be opening ourselves to questions about CUDA version compatibility, not 100% on the best path forward, but I can imagine a thought process like:
   1. I have a CUDA 12 machine
   2. I want RAPIDS
   3. Huh...only pip supports CUDA 12. I guess that means I can't use cuXFilter since it doesn't have a pip release 
   - Not sure if it makes sense to have a note that says Conda/Docker CUDA 11 installations work great on CUDA 12 machines

Contributes to #386

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -525,11 +525,6 @@
                 if (pkgs === "cucim") {
                     index_url = "";
                 }
-
-                // This should be unnecessary and removed after 23.06
-                // if (isArm) {
-                //     cupy_inst = "\npip install cupy-cuda11x -f https://pip.cupy.dev/aarch64";
-                // }
                 
                 return pip_install + pkgs + index_url + cupy_inst;
             },
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -692,7 +692,7 @@
                 }
                 if (this.active_additional_packages.includes("TensorFlow") && cuda_version !== "11.2") isDisabled = true;
                 if (this.active_os_ver === 'Ubuntu 22.04' && cuda_version !== '11.8' && this.active_method === 'Docker') isDisabled = true;
-                if (this.active_method === 'Conda' && cuda_version === '12.0') isDisabled = true;
+                if (this.active_method !== 'pip' && cuda_version === '12.0') isDisabled = true;
 
                 return isDisabled;
             },
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -516,7 +516,7 @@
                 var libraryToPkg = (pkg) => {
                     pkg = pkg.toLowerCase();
                     if (pkg === "cucim") return pkg;
-                    return pkg + cuda_suffix
+                    return pkg + cuda_suffix;
                 }
                 var pkgs = this.active_packages.map(libraryToPkg).join(" ");
                 var isArm = this.active_arch === "arm";
@@ -765,7 +765,7 @@
                     this.active_additional_packages = [];
                 }
                 if (method === 'Conda') {
-                    this.active_packages = ['Standard']
+                    this.active_packages = ['Standard'];
                 }
                 if (method === "pip") {
                     this.active_pip_cuda_ver = '12.0';
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -252,6 +252,15 @@
                         </div>
                     </template>
                 </div>
+                <div class="options-section">
+                    <div class="option-label">Method</div>
+                    <template x-for="method in methods">
+                        <div x-on:click="(e) => methodClickHandler(e, method)"
+                            x-bind:class="{'active': method === active_method}" class="option"
+                            x-html="getMethodHTML(method)">
+                        </div>
+                    </template>
+                </div>
                 <div class="options-section" x-show="active_method !== 'pip'">
                     <div class="option-label">CUDA</div>
                     <template x-for="version in cuda_vers">
@@ -266,15 +275,6 @@
                         <div x-on:click="(e) => pipCUDAClickHandler(e, version)"
                             x-bind:class="{'active': version === active_pip_cuda_ver}"
                             class="option" x-text="'CUDA ' + version"></div>
-                    </template>
-                </div>
-                <div class="options-section">
-                    <div class="option-label">Method</div>
-                    <template x-for="method in methods">
-                        <div x-on:click="(e) => methodClickHandler(e, method)"
-                            x-bind:class="{'active': method === active_method}" class="option"
-                            x-html="getMethodHTML(method)">
-                        </div>
                     </template>
                 </div>
                 <div class="options-section" x-show="active_method !== 'pip'">
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -256,13 +256,13 @@
                     <div class="option-label">Method</div>
                     <template x-for="method in methods">
                         <div x-on:click="(e) => methodClickHandler(e, method)"
-                            x-bind:class="{'active': method === active_method}" class="option"
+                            x-bind:class="{'active': method === active_method, 'disabled': disableUnsupportedMethod(method)}" class="option"
                             x-html="getMethodHTML(method)">
                         </div>
                     </template>
                 </div>
-                <div class="options-section" x-show="active_method !== 'pip'">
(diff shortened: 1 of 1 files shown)


_includes/selector.html
```diff
@@ -512,9 +527,9 @@
                 }
 
                 // This should be unnecessary and removed after 23.06
-                if (isArm) {
-                    cupy_inst = "\npip install cupy-cuda11x -f https://pip.cupy.dev/aarch64";
-                }
+                // if (isArm) {
+                //     cupy_inst = "\npip install cupy-cuda11x -f https://pip.cupy.dev/aarch64";
+                // }
```
ajschmidt8, line 532: any reason we can't just delete these lines altogether?

bdice: I think we don't need this anymore -- cupy is on pip for both AMD64 and ARM now. Safe to delete.

jarmak-nv: Yeah sorry, meant to ask if they were safe. Deleting now.

_includes/selector.html
```diff
@@ -381,16 +390,17 @@
...
             os_vers: ["Ubuntu 20.04", "Ubuntu 22.04", "CentOS 7", "Rocky Linux 8"],
             methods: ["Conda", "pip", "Docker", "Source"],
             releases: ["Stable", "Nightly"],
             arches: ["amd", "arm"],
             img_loc: ["NGC", "Dockerhub"],
-            img_types: ["core", "standard", "clx"],
+            img_types: ["core", "standard"],
             packages: ["Standard", "cuDF", "cuML", "cuGraph", "cuSpatial", "cuXFilter", "cuSignal", "cuCIM"],
-            pip_packages: ["cuDF", "cuML", "cuGraph", "cuCIM"],
+            pip_packages: ["cuDF", "cuML", "cuGraph", "cuSpatial", "cuCIM"],
```
bdice, line 403: This is waiting on https://github.com/rapidsai/cuspatial/pull/1148.

_includes/selector.html
```diff
@@ -631,9 +641,8 @@
             },
             getpipNotes() {
                 var notes = [];
-                notes = [...notes, "<code>cuDF</code>, <code>cuML</code>, and <code>cuGraph</code> pip packages are hosted on the NVIDIA Index<br>",
-                         'pip installation supports Python <code>3.8</code>, <code>3.9</code>, and <code>3.10</code><br>',
-                         'pip installation supports CUDA <code>11.2</code> - <code>11.8</code>'];
+                notes = [...notes, "<code>cuDF</code>, <code>cuML</code>, <code>cuGraph</code>, and <code>cuSpatial</code> pip packages are hosted on the NVIDIA Index<br>",
+                         'pip installation supports Python <code>3.8</code>, <code>3.9</code>, and <code>3.10</code><br>'];
```
bdice, line 640: We dropped Python 3.8 in RAPIDS 23.06. https://docs.rapids.ai/notices/rsn0029/
```suggestion
                         'pip installation supports Python <code>3.9</code> and <code>3.10</code><br>'];
```

bdice (commented):
A couple small comments/fixes.

bdice:
When do we want to merge this relative to the rollover of 23.04 -> 23.06? I think it should be concurrent, so we ought to prepare the 23.04 -> 23.06 changes in this PR and wait to merge until the release.

bdice:
Bug: Selecting `Docker` and any of (`Ubuntu 20.04`, `CentOS 7`, `Rocky Linux 8`) allows the user to choose `CUDA 12.0` but we do not support CUDA 12 for Docker.

image

_includes/selector.html
jarmak-nv: True, was thinking that saying pip supports 3.9 and 10 is a little blocking and might make users that are locked to 3.8 for whatever reason not realize they can use the wheels - however, it's not officially supported so maybe overthinking here.

jarmak-nv:
Agree we should wait until 23.06 is fully released with CUDA 12 wheels being available before merging. Will fix the docker bug - only thought to test against Conda 🤦

taureandyernv:
@jarmak-nv , if CUDA 11.x works with CUDA 12 machines, let's let the user know that for any CUDA version that we want to support.  @raydouglass @ajschmidt8 , are there any CUDA 11.x versions that we can't guarantee works in CUDA 12?

_includes/selector.html
jarmak-nv: Ah I misunderstood - ty for the clarity!

bdice: Yup, of course!

jarmak-nv:
To keep things cleaner - I'm going to spin up a new PR for the required changes to the install page itself (Python 3.9/CUDA 12 notices). 

I'll link it here for reference when it's up.

jarmak-nv:
I made a few more changes:

- Per @taureandyernv 's comment, I've made it so the `CUDA` line says different things depending on the `Method`
   - Conda/Docker show `Env. CUDA`
   - pip shows `System CUDA`
   - Intention is to differentiate them since for pip system cuda must match install cuda, but that's not the case for Docker/Conda
- I made some changes to support the selector being used Top-Down since it's kind of a frustrating experience to be moving up and down to see what's valid and what isn't
   - `Release`
      - If you select `Nightly`
         - pip is disabled
         - Docker defaults to DockerHub
      - If you select `Stable`
         - Docker returns to NGC, but the user can still manually move it back to DockerHub
            - I could go either way on this, just felt the intention was to use NGC whenever possible
   - `Method`
      - If you select `Source`
         - All options below the `Method` line are hidden since they aren't relevant to the user

ajschmidt8 (approved):

exactlyallan:
@jarmak-nv When should we pull the trigger, are we waiting on CUDA 12 like #396?

bdice (approved):
Looks fine to me. I'm unsure what merge order we need relative to #396 though.

exactlyallan:
@bdice lets just for go with #394 then #396 - do you want to do the merge honors?
@jarmak-nv thanks for all the updates

bdice:
fwiw, I would prefer to let the PR author or website PIC merge for major changes to the selector and/or releases. I would defer to someone else to merge since it looks like @exactlyallan noticed that we need changes for stable/nightly?

jarmak-nv:
For stable vs nightly version, it pulls from the website data, it's not hard defined:
https://github.com/rapidsai/docs/pull/394/files#diff-395138901913dbcb12c6504c460d16462b17ad056b0ddaca459af85680d250beR408-R413

Does https://github.com/rapidsai/rapids.ai/pull/323 make the needed changes to the site for this to work, or do we need to add it there.

On merge order, I think it's 
1. https://github.com/rapidsai/docs/pull/396
   - Update the install page so if someone finds install links during the merges they get the CUDA 12 info they need 
2. https://github.com/rapidsai/rapids.ai/pull/323
   - Update the main site so the selector has the right versions 
3. https://github.com/rapidsai/docs/pull/394
   - Finally, with everything else in place, the selector will work properly and have the documentation needed

jarmak-nv:
Added in the missing `dask-cudf` for pip installations (alongside mentioning it's on the NVIDIA Python Package Index) as well as disabled the option to use ARM + Docker + Ubuntu 20.04 [per this RSN](https://docs.rapids.ai/notices/rsn0033/).

PR #401, opened 2023-06-13
Add redirects to `stable` docs

ajschmidt8:
This PR adds a redirect for each of our libraries so that URLs like `/api/cudf/` redirect to `/api/cudf/stable/` instead of showing a 404 page.

**Note**: The `/api` pages won't actually show in PR preview environments currently.

This will be resolved as part of our planned future enhancements for our websites.


ajschmidt8:
Navigating to https://deploy-preview-401--docs-rapids-ai.netlify.app/api/cudf redirects to https://deploy-preview-401--docs-rapids-ai.netlify.app/api/cudf/stable// (note the double trailing `//`).

I think this is because there are no files in that path for PR previews.

I think we can ignore that and expect it to work correctly for the production deployment since our production deployment pulls in the documentation files from S3 (see this workflow: https://github.com/rapidsai/docs/blob/main/.github/workflows/deploy.yaml).

_redirects
```diff
@@ -0,0 +1,40 @@
+/api/cucim /api/cucim/stable/
+/api/cucim/ /api/cucim/stable/
+/api/cudf /api/cudf/stable//
```
bdice, line 3: @ajschmidt8 Was the issue you saw in this comment https://github.com/rapidsai/docs/pull/401#issuecomment-1589550633 just a typo?
```suggestion
/api/cudf /api/cudf/stable/
```

bdice (changes requested):
LGTM -- one question.

_redirects
ajschmidt8: ohhhhh shoot. yes. most likely.

i was doing bulk updates via some fancy VS Code keyboard shortcuts and they must've gone awry.

let me update this and try again.

good catch. this is why we do code reviews :slightly_smiling_face:

PR #407, opened 2023-06-21
Update `earliest` driver version for GHA Runners

ajschmidt8:
Our current `earliest` driver version, `450`, is EOL next week ([source](https://docs.nvidia.com/datacenter/tesla/drivers/#cuda-drivers)).

Therefore, we need to update to the next supported minimum driver version which is `470`.

This PR makes adjusts our docs accordingly.

### DIFF
--- a/resources/github-actions.md
+++ b/resources/github-actions.md
@@ -116,7 +116,7 @@ Operating System
 
 ### GPU Labels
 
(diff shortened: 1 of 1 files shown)


ajschmidt8:
Keeping this as a draft until the actual driver upgrades are complete.

bdice (approved):

PR #409, opened 2023-06-22
Add RMM and RAFT to install selector.

bdice:
Closes #320.


bdice:
This makes the release selector pretty "busy" with the number of packages jammed in. We may need to have further conversation on this topic with PMs / PICs. I am going to leave this as a draft for now.

exactlyallan:
Its not great, not terrible. Could break it into two rows. Or just keep the discussion going through #410

bdice:
Closing for now -- the selector changes a fair bit in #414, and it's heading in the right direction.

PR #410, opened 2023-06-23
Create "Basic" selector view

jarmak-nv:
This is a very incomplete, but potential light-weight way to make a "Basic" and "Advanced" view selector. Putting up as draft for preview + discussion.

Premise:
- I've added a top option `Basic` vs `Advanced`
- `Advanced` is what we have today
- `Basic` only gives the defaults/major RAPIDS packages + the most important option in selection (python version for conda/docker and cuda version for pip)
- This is not robust in that you can modify things in `Advanced` and going back to `Basic` doesn't reset - absolutely not ready for any kind of real review

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -235,6 +235,15 @@
         <div class="selector">
             <div class="options">
                 <div class="options-section">
+                    <div class="option-label">Selector Type</div>
+                    <template x-for="selector_type in selector_types">
+                        <div x-on:click="(e) => selectorTypeClickHandler(e, selector_type)"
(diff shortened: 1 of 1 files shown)


exactlyallan:
I like the idea - my first off the cuff thought is to move the 'basic' and 'advanced' buttons up top into tabs ([like so](https://bulma.io/documentation/components/tabs/)). Maybe replace basic with 'standard' as well.

jarmak-nv:
Closing this - 23.08 simplified the selector a lot; we can revisit if we find reason to do so.

PR #411, opened 2023-07-12
Add years to release schedule

raydouglass:
This PR does a few things to the previous releases schedule:
1. Refactors the schedule page to use `previous_releases.json` as source of truth for all versions
2. Adds the missing years for releases prior to `23.04`
3. Adds release dates for versions prior to `0.9` (derived from `cudf` Anaconda releases)

The years were calculated using this script: https://gist.github.com/raydouglass/8a0a13d06a26aca756b7d5011d7a3118

### DIFF
--- a/_data/previous_releases.json
+++ b/_data/previous_releases.json
@@ -1,5 +1,6 @@
-{
-  "2306": {
+[
+  {
+    "version": "23.06",
     "cudf_dev": {
       "start": "Mar 23 2023",
       "end": "May 17 2023",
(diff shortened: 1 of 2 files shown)


_data/previous_releases.json
```diff
@@ -35,5 +36,776 @@
...
+      "days": 4
+    },
+    "release": {
+      "start": "Jun 9 2021",
+      "end": "Jun 10 2021",
+      "days": 2
+    }
+  },
+  {
+    "version": "0.19",
```
exactlyallan, line 497: The rending of 0.19 and 0.18 are broken: 
`{“start”=>”Apr 21 2021”, “end”=>”Apr 22 2021”, “days”=>2}`



`

bdice (approved):

_data/previous_releases.json
raydouglass: Fixed!

ajschmidt8 (approved):
Approving.

In the future, I think we should consider removing all of the release metadata except for the actual release date.

I don't think people will care that the "burndown" phase of v0.15 was from Aug 6 - Aug 19, 2020 (as an example).

Including this extraneous information just means that we have to also maintain the conditional logic seen in this PR regarding things like:

- when we started splitting burndown/codefreeze into two phases for `cudf`, `rmm`, etc.
- when other repos like `rapids-cmake`, `raft`, etc joined `cudf` and `rmm` for the first phase

This list of logic will only continue to grow as we make additional changes to our processes.

Pandas, for instance, only includes the release dates in their historical list: https://pandas.pydata.org/docs/whatsnew/index.html

I couldn't find any such historical list at all for PyTorch.

Eliminating this extraneous metadata will ultimately make the site more maintainable.

ajschmidt8:
cc: @exactlyallan and @aravenel for my comment above :point_up:

PR #412, opened 2023-07-12
add cuXfilter to PyPI packages

AjayThorve:
This PR adds cuXfilter to the list of PyPI packages, since it [builds wheels](https://github.com/rapidsai/cuxfilter/pull/497) now.

I'll keep this open as a draft until the stable version of cuXfilter is released on `pypi.nvidia.com`

PR #414, opened 2023-07-21
23.08 Release Selector and Install Page Updates

### LINKED ISSUE #413: CUDA 12 on conda Install Docs/Selector updates
CUDA 12.0 is coming to conda-forge in `23.08`! We need to:

- Update the release selector for Conda
   - Adding CUDA 12.0 as an option
   - Disabling ARM as an option for CUDA 12.0 as it will not be supported in `23.08`

- Update the install page with notes about what is and is not supported in this CUDA 12.0 release
   - ARM
   - PyTorch and it's impact on cuGraph

cc. @jakirkham

jarmak-nv:
closes #413 

This PR contains the updates to the release selector required for the conda-forge CUDA 12.0 work as well as the docker overhaul.

To do:
- [x] CUDA 12 conda-forge updates
- [x] Docker Overhaul updates to the selector
- [x] Docker overhaul updates to the install page
- [x] `23.08` release updates

This PR updates:
- Selector
   - Enables CUDA 12.0 for conda packages
   - Disables CUDA 12.0 for arm conda packages
   - Disables cuSignal from conda CUDA 12.0
   - Removes the note on cuSignal for conda CUDA 12.0
   - Removes `PyCaret` as a conda additional package, it's now only available on pip for the newest versions
      - https://pypi.org/project/pycaret/#history
      - https://anaconda.org/conda-forge/pycaret/files
   - Updated to new docker setup
      - Much simpler, only `base` and `notebooks` 
      - No OS selection
      - Same CUDA 12 restrictions as Conda
 - Install page
    - Adds 2 troubleshooting steps about conda CUDA 12
       - No ARM
       - PyTorch CUDA version compatibilities
    - Expands the CUDA version compatibility section and breaks it up into docker/conda and pip 

I've also merged in #412 to simplify merging at release.

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -333,6 +333,7 @@
                                 class="option" x-text="type"></div>
                         </template>
                     </div>
+                </div>
                 <div class="note" x-show="getNoteHtml()">
                     <div class="option-label"></div>
                     <div class="option-blank">
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -536,7 +536,6 @@
 
                 var imgRepo = [
                     imgLoc + "rapidsai/" + imgVariant,
-                    (isNightly ? "nightly" : ""),
                 ].filter(Boolean).join("-");
 
                 var imgTag = [
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -247,15 +247,6 @@
                         </div>
                     </template>
                 </div>
-                <!-- <div class="options-section">
-                    <div class="option-label">Arch</div>
-                    <template x-for="arch in arches">
-                        <div x-on:click="(e) => archClickHandler(e, arch)"
-                            x-bind:class="{'active': arch === active_arch, 'disabled': disableUnsupportedArch(arch)}" class="option" x-text="arch"
-                            x-text="getArchText(arch)">
-                        </div>
-                    </template>
-                </div> -->
                 <div class="options-section">
                     <div class="option-label">Method</div>
                     <template x-for="method in methods">
@@ -384,7 +375,6 @@
             active_pip_cuda_ver: "12.0",
             active_method: "Conda",
             active_release: "Stable",
-            // active_arch: "amd",
             active_img_type: "Base",
             active_img_loc: "NGC",
             active_packages: ["Standard"],
@@ -396,7 +386,6 @@
             pip_cuda_vers: ["11.2 - 11.8", "12.0"],
             methods: ["Conda", "pip", "Docker"],
             releases: ["Stable", "Nightly"],
-            // arches: ["amd", "arm"],
             img_loc: ["NGC", "Dockerhub"],
             img_types: ["Base", "Notebooks"],
             packages: ["Standard", "Choose Specific Packages"],
@@ -414,7 +403,6 @@
             getMethodHTML(method) {
                 var icon_class = "box-open fas";
                 if (method.includes("Docker")) icon_class = "regular fa-container-storage";
-                // if (method.includes("Source")) icon_class = "cogs fas";
                 return method + "&nbsp;<i class='fa-" + icon_class + "'></i>";
             },
             getPackageHTML(pkg) {
@@ -430,12 +418,6 @@
                 if (release.includes("Nightly")) version = this.getNightlyVersion() + "a";
                 return release + " " + "(" + version + ")";
             },
-            // getArchText(arch) {
-            //     var text = "Arch not implemented yet!"
-            //     if (arch === "amd") text = "x86-64";
-            //     if (arch === "arm") text = "Arm Server (SBSA)";
-            //     return text;
-            // },
             highlightCmd(str) {
                 return "<span class='nb'>" + str + "</span>"
             },
@@ -520,7 +502,6 @@
                         .filter(pkg => pkg !== "Choose Specific Packages" && pkg !== "cuSignal")
                         .map(libraryToPkg)
                         .join(" ");                }
-                // var isArm = this.active_arch === "arm";
 
                 if (pkgs === "cucim") {
                     index_url = "";
@@ -559,29 +540,6 @@
                     indent + fullImg;
                 return cmd;
             },
-            // getSourceCmdHtml() {
-            //     var isStable = this.active_release === "Stable";
-            //     var version = isStable ? this.getStableVersion() : this.getNightlyVersion();
-            //     var links = [
-            //         "https://github.com/rapidsai/cudf/blob/branch-" + version + "/CONTRIBUTING.md",
-            //         "https://github.com/rapidsai/cuml/blob/branch-" + version + "/README.md",
-            //         "https://github.com/rapidsai/cugraph/blob/branch-" + version + "/readme_pages/CONTRIBUTING.md",
-            //         "https://github.com/rapidsai/cuspatial/blob/branch-" + version + "/README.md",
-            //         "https://github.com/rapidsai/cuxfilter/blob/branch-" + version + "/CONTRIBUTING.md",
-            //         "https://github.com/rapidsai/cusignal#source-linux-os",
-            //         "https://github.com/rapidsai/cucim/blob/branch-" + version + "/CONTRIBUTING.md#building-and-testing-cucim-from-source"
-            //     ];
-            //     var cmd = "<p class='selector__source-cmd'>For source build instructions, please view the links below:</p>";
-            //     cmd += "<ul class='selector__source-cmd__list'>";
-            //     for (var i = 0; i < links.length; i++) {
-            //         var link = links[i];
-            //         cmd += "<li class='selector__source-cmd__list-item'>";
-            //         cmd += "<a class='selector__source-cmd__link' target='_blank' href='" + link + "'>" + link + "</a>";
-            //         cmd += "</li>";
-            //     }
-            //     cmd += "</ul>";
-            //     return cmd;
-            // },
             getImgLoc(loc) {
                 if (loc === "NGC") return 'NGC (Stable Only) <i class="fa-regular fa-share-nodes"></i>';
                 if (loc === "Dockerhub") return 'Docker Hub (Stable and Nightly) <i class="fa-docker fab"></i>';
@@ -628,15 +586,8 @@
 
                 return notes.map(note => this.note_prefix + " " + note);
             },
-            // getArmNotes() {
-            //     return [this.note_prefix + " ARM support does not currently include Jetson products."];
-            // },
             getNoteHtml() {
                 var notes = [];
-
-                // if (this.active_arch === "arm") {
-                //     notes.push(...this.getArmNotes());
-                // }
 
                 if (this.active_method === "Docker") {
                     notes.push(...this.getDockerNotes());
@@ -659,8 +610,6 @@
                     return this.getCondaCmdHtml();
                 if (this.active_method === "pip")
                     return this.getpipCmdHtml();
-                // if (this.active_method === "Source")
-                //     return this.getSourceCmdHtml();
                 return "Not implemented yet!";
             },
             disableUnsupportedRelease(release) {
@@ -672,8 +621,6 @@
             disableUnsupportedCuda(cuda_version) {
                 var isDisabled = false;
                 if (this.active_additional_packages.includes("TensorFlow") && cuda_version !== "11.2") isDisabled = true;
-                // if (this.active_arch === "arm" && cuda_version === "12.0") isDisabled = true;
-
                 return isDisabled;
             },
             disableUnsupportedPython(python_version) {
@@ -693,11 +640,6 @@
                 if (this.active_release === "Nightly" && method === "pip") isDisabled = true;
                 return isDisabled;
             },
-            // disableUnsupportedArch(arch) {
-            //     var isDisabled = false;
-            //     if (["Conda", "Docker"].includes(this.active_method) && this.active_cuda_ver.startsWith("12") && arch === "arm") isDisabled = true;
-            //     return isDisabled;
-            // },
             disableUnsupportedPackage(package) {
                 var isDisabled = false;
                 if (this.active_cuda_ver.startsWith("12") && this.active_method === "Conda" && package === "cuSignal") isDisabled = true;
@@ -717,10 +659,6 @@
                 if (this.isDisabled(e.target)) return;
                 this.active_img_type = type;
             },
-            // archClickHandler(e, arch) {
-            //     if (this.isDisabled(e.target)) return;
-            //     this.active_arch = arch;
-            // },
             cudaClickHandler(e, version) {
                 if (this.isDisabled(e.target)) return;
                 this.active_cuda_ver = version;
@@ -748,7 +686,6 @@
                 if (method === "pip") {
                     this.active_pip_cuda_ver = '12.0';
                     this.active_release = 'Stable';
-                    // this.active_packages = ['cuDF']
                 }
                 if (method === "Docker") {
                     if (this.active_release === 'Nightly') {
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -377,8 +377,8 @@
             img_loc: ["NGC", "Dockerhub"],
             img_types: ["Base", "Notebooks"],
             packages: ["Standard", "Choose Specific Packages"],
-            additional_pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial + cuProj", "cuxfilter", "cuCIM"],
-            additional_rapids_packages: ["cuDF", "cuML", "cuGraph", "cuSpatial + cuProj", "cuxfilter", "cuSignal", "cuCIM"],
+            additional_pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuCIM"],
+            additional_rapids_packages: ["cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuSignal", "cuCIM"],
             additional_packages: ["Dask-SQL", "JupyterLab", "Plotly Dash", "Graphistry", "TensorFlow", "Xarray-Spatial"],
             note_prefix: "<i class='fas fa-info-circle text-blue'></i>",
             rapids_meta_pkgs: ["cuDF", "cuML", "cuGraph", "cuSpatial", "cuProj", "cuxfilter", "cuSignal", "cuCIM"],
@@ -452,8 +452,8 @@
                     pkgs = ["rapids"];
                 }
 
-                if (pkgs.includes("cuSpatial + cuProj")) {
-                    pkgs = pkgs.filter(pkg => pkg !== "cuSpatial + cuProj");
+                if (pkgs.includes("cuSpatial/cuProj")) {
+                    pkgs = pkgs.filter(pkg => pkg !== "cuSpatial/cuProj");
                     pkgs.push("cuSpatial");
                     pkgs.push("cuProj");
                 }
@@ -492,7 +492,7 @@
                 var libraryToPkg = (pkg) => {
                     pkg = pkg.toLowerCase();
                     if (pkg === "cucim") return pkg;
-                    if (pkg == "cuspatial + cuproj") return "cuspatial" + cuda_suffix + " cuproj" + cuda_suffix;
+                    if (pkg == "cuspatial/cuproj") return "cuspatial" + cuda_suffix + " cuproj" + cuda_suffix;
                     return pkg + cuda_suffix;
                 }
                 if (this.active_packages.length === 1 && this.active_packages[0] === "Choose Specific Packages") {
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -18,6 +18,14 @@
         flex-wrap: nowrap;
     }
 
+    .selector .options-section-specific {
+        display: flex;
+        flex-direction: row;
+        flex-wrap: nowrap;
+        width: 97%;
+        margin: 0 auto;
+    }
+
     .selector .options .option-label {
         color: white;
         width: 6em;
@@ -290,8 +298,8 @@
                             </div>
                         </template>
                     </div>
-                    <div class="options-section" x-show="active_packages != 'Standard'">
-                        <div class="option-label">RAPIDS Packages</div>
+                    <div class="options-section-specific" x-show="active_packages != 'Standard'">
+                        <div class="option-label"><!-- Second row of packages --></div>
                         <template x-for="package in additional_rapids_packages">
                             <div x-on:click="(e) => packagesClickHandler(e, package)"
                                 x-bind:class="{'active': active_packages.includes(package)}" class="option"
@@ -317,7 +325,7 @@
                             </div>
                         </template>
                     </div>
-                    <div class="options-section" x-show="active_packages != 'Standard'">
+                    <div class="options-section-specific" x-show="active_packages != 'Standard'">
                         <div class="option-label"><!-- Second row of packages --></div>
                         <template x-for="package in additional_pip_packages">
                             <div x-on:click="(e) => packagesClickHandler(e, package)"
@@ -709,6 +717,15 @@
                 if (package === "Standard") {
                     this.active_packages = [package];
                     return;
+                }
+
+                if (package === "Choose Specific Packages") {
+                    if (this.active_packages[0] === "Standard" && this.active_packages.length === 1) {
+                        this.active_packages = [package, "cuDF"];
+                        return;
+                    } else {
+                        return;
+                    }
                 }
 
                 this.active_packages = this.active_packages.filter(el => el !== "Standard");
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -560,6 +560,7 @@
             },
             getDockerNotes() {
                 var cuda12 = ""
+                var notes = [];
                 if (this.active_cuda_ver.startsWith("12")) {
                         var pkgs_html = this.rapids_meta_pkgs.filter(pkg => pkg !== "cuSignal").map(pkg => "<code>" + pkg + "</code>").join(", ");
                         cuda12 = "RAPIDS Docker images for CUDA 12 do not currently support ARM architectures<br>";
@@ -567,7 +568,10 @@
                         var pkgs_html = this.rapids_meta_pkgs.map(pkg => "<code>" + pkg + "</code>").join(", ");
                     }
 
-                return [this.note_prefix + cuda12 + " The selected image contains the following packages:<br>" + pkgs_html];
+                // return [this.note_prefix + cuda12 + " The selected image contains the following packages:<br>" + pkgs_html];
+                notes = [cuda12, "The selected image contains the following packages:<br>", pkgs_html]
+                return notes.map(note => this.note_prefix + " " + note);
+
             },
             getCondaNotes() {
                 var notes = [];
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -326,31 +326,13 @@
                         </template>
                     </div>
                     <div class="options-section">
-                        <div class="option-label">Image OS</div>
-                        <template x-for="version in os_vers">
-                            <div x-on:click="(e) => imgOSClickHandler(e, version)"
-                                x-bind:class="{'active': version === active_os_ver, 'disabled': disableUnsupportedImgOS(version)}"
-                                class="option" x-html="getOSHTML(version)">
-                            </div>
-                        </template>
-                    </div>
-                    <div class="options-section">
                         <div class="option-label">Image Type</div>
                         <template x-for="type in img_types">
                             <div x-on:click="(e) => imgTypeClickHandler(e, type)"
                                 x-bind:class="{'active': type === active_img_type, 'disabled': disableUnsupportedImgType(type)}"
-                                class="option" x-text="getImgTypeText(type)"></div>
+                                class="option" x-text="type"></div>
                         </template>
                     </div>
-                    <div class="options-section">
-                        <div class="option-label">Image Options</div>
-                        <template x-for="option in img_options">
-                            <div x-on:click="(e) => imgOptionClickHandler(e, option)"
-                                x-bind:class="{'active': active_img_options.includes(option), 'disabled': disableUnsupportedImgOption(option)}"
-                                class="option" x-text="getImgOptionText(option)"></div>
-                        </template>
-                    </div>
-                </div>
                 <div class="note" x-show="getNoteHtml()">
                     <div class="option-label"></div>
                     <div class="option-blank">
@@ -379,30 +361,26 @@
             active_python_ver: "3.10",
             active_cuda_ver: "11.8",
             active_pip_cuda_ver: "12.0",
-            active_os_ver: "Ubuntu 22.04",
             active_method: "Conda",
             active_release: "Stable",
             active_arch: "amd",
-            active_img_type: "core",
+            active_img_type: "Base",
             active_img_loc: "NGC",
             active_packages: ["Standard"],
-            active_img_options: ["notebooks"],
             active_additional_packages: [],
 
             // all possible values
             python_vers: ["3.9", "3.10"],
             cuda_vers: ["11.2", "11.4", "11.5", "11.8", "12.0"],
             pip_cuda_vers: ["11.2 - 11.8", "12.0"],
-            os_vers: ["Ubuntu 20.04", "Ubuntu 22.04", "CentOS 7", "Rocky Linux 8"],
             methods: ["Conda", "pip", "Docker", "Source"],
             releases: ["Stable", "Nightly"],
             arches: ["amd", "arm"],
             img_loc: ["NGC", "Dockerhub"],
-            img_types: ["core", "standard"],
+            img_types: ["Base", "Notebook"],
             packages: ["Standard", "cuDF", "cuML", "cuGraph", "cuSpatial", "cuXFilter", "cuSignal", "cuCIM"],
             pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial", "cuCIM"],
             additional_packages: ["Dask SQL", "JupyterLab", "Plotly Dash", "Graphistry", "TensorFlow", "Xarray-Spatial"], // , "RTXpy" will return later
-            img_options: ["notebooks", "development"],
             note_prefix: "<i class='fas fa-info-circle text-blue'></i>",
             rapids_meta_pkgs: ["cuDF", "cuML", "cuGraph", "cuSpatial", "cuXFilter", "cuSignal", "cuCIM"],
             getStableVersion() {
@@ -410,12 +388,6 @@
             },
             getNightlyVersion() {
                 return "{{ site.data.releases.nightly.version }}";
-            },
-            getOSHTML(os_ver) {
-                var icon_class = "ubuntu"
-                if (os_ver.includes("CentOS")) icon_class = "centos";
-                if (os_ver.includes("Rocky Linux")) icon_class = "linux";
-                return os_ver + "&nbsp;<i class='fab fa-" + icon_class + "'></i>";
             },
             getMethodHTML(method) {
                 var icon_class = "box-open fas";
@@ -529,33 +501,24 @@
                 return pip_install + pkgs + index_url + cupy_inst;
             },
             getDockerCmdHtml() {
-                var hasNotebooks = this.active_img_options.includes("notebooks");
-                var isDevel = this.active_img_options.includes("development");
+                var hasNotebooks = this.active_img_type === "Notebook";
                 var isNightly = this.active_release === "Nightly";
-                var isArm = this.active_arch === "arm";
                 var imgLoc = "";
-                if (this.active_img_loc === "NGC") imgLoc = "nvcr.io/nvidia/"
-                var imgVariant = "";
-                if (this.active_img_type === "core") imgVariant = "core";
-                var imgType = "base";
-                if (hasNotebooks) imgType = "runtime";
-                if (isDevel) imgType = "devel";
+                if (this.active_img_loc === "NGC") imgLoc = "nvcr.io/nvidia/";
+                var imgVariant = this.active_img_type;
+
                 var portOptions = (hasNotebooks ? ["8888:8888", "8787:8787", "8786:8786"] : [])
                     .map(opt => this.highlightFlag("-p") + " " + opt + " ").join("");
 
                 var imgRepo = [
                     imgLoc + "rapidsai/rapidsai",
                     imgVariant,
-                    (isDevel ? "dev" : ""),
                     (isNightly ? "nightly" : ""),
-                    (isArm ? "arm64" : ""),
                 ].filter(Boolean).join("-");
 
                 var imgTag = [
                     (isNightly ? this.getNightlyVersion() : this.getStableVersion()),
                     "cuda" + this.active_cuda_ver,
-                    imgType,
-                    this.active_os_ver.replaceAll(" ", ""),
                     "py" + this.active_python_ver
                 ].join("-");
 
@@ -591,16 +554,6 @@
                 }
                 cmd += "</ul>";
                 return cmd;
-            },
-            getImgTypeText(type) {
-                if (type === "core") return "Basic"
-                if (type === "standard") return "Basic w/ Dask-SQL"
-                return type;
-            },
-            getImgOptionText(option) {
-                if (option === "notebooks") return "Include Jupyter Notebooks";
-                if (option === "development") return "Development Image (only on Docker Hub)";
-                return option;
             },
             getImgLoc(loc) {
                 if (loc === "NGC") return 'NGC (Stable Only) <i class="fa-regular fa-share-nodes"></i>';
@@ -694,11 +647,7 @@
             },
             disableUnsupportedCuda(cuda_version) {
                 var isDisabled = false;
-                if (this.active_img_options.includes("development") && cuda_version !== this.getNewestCudaVer()) {
-                    isDisabled = true;
-                }
                 if (this.active_additional_packages.includes("TensorFlow") && cuda_version !== "11.2") isDisabled = true;
-                if (this.active_os_ver === 'Ubuntu 22.04' && cuda_version !== '11.8' && this.active_method === 'Docker') isDisabled = true;
                 if (this.active_arch === "arm" && cuda_version === "12.0") isDisabled = true;
 
                 return isDisabled;
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -372,7 +372,7 @@
 
             // all possible values
             python_vers: ["3.9", "3.10"],
-            cuda_vers: ["11.2", "11.4", "11.5", "11.8", "12.0"],
+            cuda_vers: ["11.2", "11.8", "12.0"],
             pip_cuda_vers: ["11.2 - 11.8", "12.0"],
             methods: ["Conda", "pip", "Docker", "Source"],
             releases: ["Stable", "Nightly"],
@@ -506,15 +506,14 @@
                 var isNightly = this.active_release === "Nightly";
                 var imgLoc = "";
                 if (this.active_img_loc === "NGC") imgLoc = "nvcr.io/nvidia/";
-                var imgVariant = "";
+                var imgVariant = "Base";
                 if (this.active_img_type === "Notebooks") imgVariant = this.active_img_type;
 
                 var portOptions = (hasNotebooks ? ["8888:8888", "8787:8787", "8786:8786"] : [])
                     .map(opt => this.highlightFlag("-p") + " " + opt + " ").join("");
 
                 var imgRepo = [
-                    imgLoc + "rapidsai/rapidsai",
-                    imgVariant,
+                    imgLoc + "rapidsai/" + imgVariant,
                     (isNightly ? "nightly" : ""),
                 ].filter(Boolean).join("-");
 
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -582,7 +582,7 @@
             getpipNotes() {
                 var notes = [];
                 notes = [...notes, "<code>cuDF</code>, <code>dask-cuDF</code>, <code>cuML</code>, <code>cuGraph</code>, <code>cuSpatial</code>, and <code>cuXfilter</code> pip packages are hosted on the NVIDIA Python Package Index<br>",
-                    'pip installation supports Python <code>3.9</code>, and <code>3.10</code><br>'];
+                    'pip installation supports Python <code>3.9</code> and <code>3.10</code><br>'];
 
                 return notes.map(note => this.note_prefix + " " + note);
             },
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -9,6 +9,7 @@
     .selector-bg {
         background-color: #9943ff;
         padding: 1rem;
+        border-radius: 5px;
     }
 
     .selector .options-section {
@@ -36,6 +37,7 @@
         cursor: pointer;
         line-height: 1.5em;
         box-shadow: 2px 2px 2px rgba(10, 10, 10, 0.4);
+        border-radius: 2px;
     }
 
     .selector .options .option:hover,
@@ -61,6 +63,7 @@
         background: #e3e3e3;
         color: #000000;
         padding: 0.1em 0.1em 0.15em;
+        border-radius: 2px;
     }
 
     .selector .options .option-blank {
@@ -512,9 +515,10 @@
                 if (this.active_packages[0] === 'Standard') {
                     var pkgs = this.additional_pip_packages.map(libraryToPkg).join(" ");
                 } else {
-                    var pkgs = this.active_packages.filter(pkg => pkg !== "Choose Specific Packages").map(libraryToPkg).join(" ");
-                }
-                // var pkgs = this.active_packages.filter(pkg => pkg !== "Choose Specific Packages").map(libraryToPkg).join(" ");
+                    var pkgs = this.active_packages
+                        .filter(pkg => pkg !== "Choose Specific Packages" && pkg !== "cuSignal")
+                        .map(libraryToPkg)
+                        .join(" ");                }
                 // var isArm = this.active_arch === "arm";
 
                 if (pkgs === "cucim") {
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -563,10 +563,9 @@
                 return loc;
             },
             getDockerNotes() {
-                if (this.active_cuda_ver === "12.0") {
+                if (this.active_cuda_ver.startsWith("12")) {
                         var pkgs_html = this.rapids_meta_pkgs.filter(pkg => pkg !== "cuSignal").map(pkg => "<code>" + pkg + "</code>").join(", ");
-                    }
-                else {
+                } else {
                         var pkgs_html = this.rapids_meta_pkgs.map(pkg => "<code>" + pkg + "</code>").join(", ");
                     }
 
@@ -576,10 +575,9 @@
                 var notes = [];
                 notes = [...notes, "RAPIDS currently doesn't support <code>channel_priority: strict</code>; use <code>channel_priority: flexible</code> instead"];
                 if (this.active_packages.length === 1 && this.active_packages[0] === "Standard") {
-                    if (this.active_cuda_ver === "12.0") {
+                    if (this.active_cuda_ver.startsWith("12")) {
                         var pkgs_html = this.rapids_meta_pkgs.filter(pkg => pkg !== "cuSignal").map(pkg => "<code>" + pkg + "</code>").join(", ");
-                    }
-                    else {
+                    } else {
                         var pkgs_html = this.rapids_meta_pkgs.map(pkg => "<code>" + pkg + "</code>").join(", ");
                     }
                     notes = [...notes, "The CUDA " + this.active_cuda_ver + 
@@ -671,13 +669,12 @@
             },
             disableUnsupportedArch(arch) {
                 var isDisabled = false;
-                if (this.active_method === "Conda" && this.active_cuda_ver === "12.0" && arch === "arm") isDisabled = true;
-                if (this.active_method === "Docker" && this.active_cuda_ver === "12.0" && arch === "arm") isDisabled = true;
+                if (["Conda", "Docker"].includes(this.active_method) && this.active_cuda_ver.startsWith("12") && arch === "arm") isDisabled = true;
                 return isDisabled;
             },
             disableUnsupportedPackage(package) {
                 var isDisabled = false;
-                if (this.active_cuda_ver === "12.0" && this.active_method === "Conda" && package === "cuSignal") isDisabled = true;
+                if (this.active_cuda_ver.startsWith("12") && this.active_method === "Conda" && package === "cuSignal") isDisabled = true;
                 return isDisabled;
             },
             isDisabled(e) {
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -261,7 +261,7 @@
                         </div>
                     </template>
                 </div>
-                <div class="options-section" x-show="active_method !== 'pip' && active_method !== 'Source'">
+                <div class="options-section" x-show="active_method !== 'pip'">
                     <div class="option-label">ENV. CUDA</div>
                     <template x-for="version in cuda_vers">
                         <div x-on:click="(e) => cudaClickHandler(e, version)"
@@ -277,7 +277,7 @@
                             class="option" x-text="'CUDA ' + version"></div>
                     </template>
                 </div>
-                <div class="options-section" x-show="active_method !== 'pip' && active_method !== 'Source'">
+                <div class="options-section" x-show="active_method !== 'pip'">
                     <div class="option-label">Python</div>
                     <template x-for="version in python_vers">
                         <div x-on:click="(e) => pythonClickHandler(e, version)"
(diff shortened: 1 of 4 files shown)


jarmak-nv:
I think this is good to review. I was initially holding back to consider adding the docker changes to the install page, but I'll do that in a separate PR to open this to review now.

Selector changes _are_ updated for Docker since CUDA 12 update impacts that as well.

jarmak-nv:
@raydouglass would also appreciate your eyes on the selector when you find a moment to make sure I set up the Docker changes correctly - thanks! (I can't add you as a reviewer otherwise I'd just do that.)

_includes/selector.html
```diff
@@ -379,30 +362,26 @@
             active_python_ver: "3.10",
             active_cuda_ver: "11.8",
```
bdice, line 363: Should we default to 12.0 even though it's not available for ARM?
```suggestion
            active_cuda_ver: "12.0",
```

_includes/selector.html
```diff
@@ -379,30 +362,26 @@
...
             releases: ["Stable", "Nightly"],
             arches: ["amd", "arm"],
             img_loc: ["NGC", "Dockerhub"],
-            img_types: ["core", "standard"],
+            img_types: ["Base", "Notebooks"],
             packages: ["Standard", "cuDF", "cuML", "cuGraph", "cuSpatial", "cuXFilter", "cuSignal", "cuCIM"],
             pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial", "cuCIM"],
-            additional_packages: ["Dask SQL", "JupyterLab", "Plotly Dash", "Graphistry", "PyCaret", "TensorFlow", "Xarray-Spatial"], // , "RTXpy" will return later
-            img_options: ["notebooks", "development"],
+            additional_packages: ["Dask SQL", "JupyterLab", "Plotly Dash", "Graphistry", "TensorFlow", "Xarray-Spatial"], // , "RTXpy" will return later
```
bdice, line 384: Rename `Dask SQL` to `Dask-SQL` (or `dask-sql`) anywhere that occurs.
```suggestion
            additional_packages: ["Dask-SQL", "JupyterLab", "Plotly Dash", "Graphistry", "TensorFlow", "Xarray-Spatial"], // , "RTXpy" will return later
```

_includes/selector.html
```diff
@@ -592,36 +557,33 @@
...
                 if (loc === "NGC") return 'NGC (Stable Only) <i class="fa-regular fa-share-nodes"></i>';
                 if (loc === "Dockerhub") return 'Docker Hub (Stable and Nightly) <i class="fa-docker fab"></i>';
                 return loc;
             },
             getDockerNotes() {
-                var core_pkgs = this.rapids_meta_pkgs;
+                if (this.active_cuda_ver === "12.0") {
+                        var pkgs_html = this.rapids_meta_pkgs.filter(pkg => pkg !== "cuSignal").map(pkg => "<code>" + pkg + "</code>").join(", ");
+                    }
+                else {
```
bdice, line 569:
```suggestion
                } else {
```

_includes/selector.html
```diff
@@ -592,36 +557,33 @@
...
-                return option;
-            },
             getImgLoc(loc) {
                 if (loc === "NGC") return 'NGC (Stable Only) <i class="fa-regular fa-share-nodes"></i>';
                 if (loc === "Dockerhub") return 'Docker Hub (Stable and Nightly) <i class="fa-docker fab"></i>';
                 return loc;
             },
             getDockerNotes() {
-                var core_pkgs = this.rapids_meta_pkgs;
+                if (this.active_cuda_ver === "12.0") {
```
bdice, line 566: Maybe write this more defensively for versions like 12.1?
```suggestion
                if (this.active_cuda_ver.startsWith("12")) {
```

_includes/selector.html
```diff
@@ -592,36 +557,33 @@
...
-                var pkgs_html = core_pkgs.map(pkg => "<code>" + pkg + "</code>").join(", ");
                 return [this.note_prefix + " The selected image contains the following packages:<br>" + pkgs_html];
             },
             getCondaNotes() {
                 var notes = [];
                 notes = [...notes, "RAPIDS currently doesn't support <code>channel_priority: strict</code>; use <code>channel_priority: flexible</code> instead"];
                 if (this.active_packages.length === 1 && this.active_packages[0] === "Standard") {
-                    var pkgs_html = this.rapids_meta_pkgs.map(pkg => "<code>" + pkg + "</code>").join(", ");
-                    notes = [...notes, "The <code>Standard</code> selection contains the following packages:<br>" + pkgs_html];
+                    if (this.active_cuda_ver === "12.0") {
```
bdice, line 579: Maybe write this more defensively for versions like 12.1?
```suggestion
                    if (this.active_cuda_ver.startsWith("12")) {
```

_includes/selector.html
```diff
@@ -592,36 +557,33 @@
...
+                        var pkgs_html = this.rapids_meta_pkgs.filter(pkg => pkg !== "cuSignal").map(pkg => "<code>" + pkg + "</code>").join(", ");
+                    }
+                    else {
```
bdice, line 582:
```suggestion
                    } else {
```

_includes/selector.html
```diff
@@ -688,44 +646,38 @@
...
                 var isDisabled = false;
                 return isDisabled;
             },
-            disableUnsupportedImgOS(os) {
+            disableUnsupportedImgType(type) {
                 var isDisabled = false;
-                if (this.active_arch === "arm" && os === "CentOS 7") isDisabled = true;
-                if (this.active_cuda_ver !== '11.8' && os === 'Ubuntu 22.04' && this.active_method === 'Docker') isDisabled = true;
-                if (this.active_arch === "arm" && os === 'Ubuntu 20.04' && this.active_method === 'Docker') isDisabled = true;
+                if (this.active_arch === "arm" && type !== "core") isDisabled = true;
```
bdice, line 660: This should be checking either `"Base"` or `"Notebooks"` and not `"core"`, right? I'm not entirely sure what this logic is meant to be doing -- do we support both base/notebooks with ARM?

_includes/selector.html
```diff
@@ -688,44 +646,38 @@
...
-                var isNGC = this.active_img_loc === "NGC"
-                if (isNGC && option === "development") isDisabled = true;
+                if (this.active_release === "Nightly" && method === "pip") isDisabled = true;
                 return isDisabled;
             },
-            disableUnsupportedImgLoc(loc) {
+            disableUnsupportedArch(arch) {
                 var isDisabled = false;
+                if (this.active_method === "Conda" && this.active_cuda_ver === "12.0" && arch === "arm") isDisabled = true;
+                if (this.active_method === "Docker" && this.active_cuda_ver === "12.0" && arch === "arm") isDisabled = true;
```
bdice, line 675: Condense this logic a bit. Conda/Docker can be handled together. Either suggestion below is okay.

```suggestion
                if (["Conda", "Docker"].includes(this.active_method) && this.active_cuda_ver.startsWith("12") && arch === "arm") isDisabled = true;
```

```suggestion
                if ((this.active_method === "Conda" || this.active_method === "Docker") && this.active_cuda_ver.startsWith("12") && arch === "arm") isDisabled = true;
```

_includes/selector.html
```diff
@@ -688,44 +646,38 @@
...
                 return isDisabled;
             },
-            disableUnsupportedMethod(method) {
+            disableUnsupportedPackage(package) {
                 var isDisabled = false;
-                if (this.active_release === "Nightly" && method === "pip") isDisabled = true;
+                if (this.active_cuda_ver === "12.0" && this.active_method === "Conda" && package === "cuSignal") isDisabled = true;
```
bdice, line 680:
```suggestion
                if (this.active_cuda_ver.startsWith("12") && this.active_method === "Conda" && package === "cuSignal") isDisabled = true;
```

install/install.md
```diff
@@ -47,6 +47,16 @@ Use the selector tool below to select your preferred method, packages, and envir
 ## Installation Troubleshooting
 
 ### **Conda Issues**
+<i class="fas fa-info-circle"></i> CUDA 12.0 arm packages don't exist:<br/>
```
bdice, line 50:
```suggestion
<i class="fas fa-info-circle"></i> CUDA 12.0 ARM packages are not yet available:<br/>
```

install/install.md
```diff
@@ -47,6 +47,16 @@ Use the selector tool below to select your preferred method, packages, and envir
...
+Conda-forge doesn't yet support the minimum required glibc (2.27) for CUDA 12 on arm. For arm support, please use an 11.x installation.
```
bdice, line 51:
```suggestion
Conda-forge does not yet support the minimum required glibc (2.27) for CUDA 12 on ARM. For ARM support, please use CUDA 11.
```

install/install.md
```diff
@@ -47,6 +47,16 @@ Use the selector tool below to select your preferred method, packages, and envir
...
+
+<i class="fas fa-info-circle"></i> CUDA 12.0 packages don't work with stable PyTorch packages: <br/>
+PyTorch currently only has nightly builds for CUDA 12.1, stable builds are limited to CUDA 11.x. <br/>
```
bdice, line 54:
```suggestion
PyTorch currently only has nightly builds for CUDA 12.1, stable builds are limited to CUDA 11. <br/>
```

install/install.md
```diff
@@ -47,6 +47,16 @@ Use the selector tool below to select your preferred method, packages, and envir
 ## Installation Troubleshooting
 
 ### **Conda Issues**
+<i class="fas fa-info-circle"></i> CUDA 12.0 arm packages don't exist:<br/>
+Conda-forge doesn't yet support the minimum required glibc (2.27) for CUDA 12 on arm. For arm support, please use an 11.x installation.
+
+<i class="fas fa-info-circle"></i> CUDA 12.0 packages don't work with stable PyTorch packages: <br/>
```
bdice, line 53: I'd be a little more defensive here because this could change at any time.
```suggestion
<i class="fas fa-info-circle"></i> At the time of writing, there is no stable CUDA 12 release of PyTorch: <br/>
```

install/install.md
```diff
@@ -47,6 +47,16 @@ Use the selector tool below to select your preferred method, packages, and envir
...
+PyTorch currently only has nightly builds for CUDA 12.1, stable builds are limited to CUDA 11.x. <br/>
+The installation method below may function if there is a hard-requirement of CUDA 12 but it is currently unsupported:
```
bdice, line 55:
```suggestion
The installation method below may allow RAPIDS CUDA 12.0 packages to coexist with PyTorch CUDA 12.1 nightly packages if there is a hard-requirement of CUDA 12 but it is currently unsupported:
```

install/install.md
```diff
@@ -47,6 +47,16 @@ Use the selector tool below to select your preferred method, packages, and envir
...
+```
+mamba create -n rapids-pytorch-cu12 -c rapidsai -c pytorch-nightly -c conda-forge -c nvidia cuml pytorch pytorch-cuda=12.1 cuda-version=12.0
```
bdice, line 57: We want to install all of RAPIDS and not just cuML, right? Also do we want a version pinning to try and ensure that things are fetched properly?
```suggestion
mamba create -n rapids-pytorch-cu12 -c rapidsai -c pytorch-nightly -c conda-forge -c nvidia rapids=23.08 cuda-version=12.0 pytorch pytorch-cuda=12.1
```

install/install.md
```diff
@@ -119,11 +129,22 @@ All provisioned systems need to be RAPIDS capable. Here's what is required:
...
 
  **Note**: RAPIDS is tested with and officially supports the versions listed above. Newer CUDA and driver versions may also work with RAPIDS. See [CUDA compatibility](https://docs.nvidia.com/deploy/cuda-compatibility/index.html) for details. 
+
+## **CUDA 12 Support**
+
+### **Docker and Conda**
+
+- <i class="fas fa-info-circle"></i> CUDA 12 packages and images currently support CUDA 12.0
+- <i class="fas fa-info-circle"></i> CUDA 11 packages and images can be used on a system with a CUDA 12 driver because they include their own CUDA toolkit
+- <i class="fas fa-info-circle"></i> arm is not currently supported on CUDA 12 packages and images, use CUDA 11 or pip for arm support
```
bdice, line 142:
```suggestion
- <i class="fas fa-info-circle"></i> ARM is not currently supported on CUDA 12 packages and images, use CUDA 11 or pip packages for ARM support
```

bdice (commented):
Comments attached.

_includes/selector.html
mmccarty: I would think we would wait until 12 is supported on all platforms.

jarmak-nv: Agree with Mike, I should have commented that in the code instead of in the comment above.

install/install.md
jarmak-nv: ARM capitalization is confusing! Happy to get a clear answer here lol (Company name is `Arm`, logo is `arm` and arch is `ARM` it seems!)

_includes/selector.html
jarmak-nv: I like this! In `23.10` I think this logic isn't necessary anyway since we won't be including cuSignal anywhere, right? But it's good practice and I'll add it in.

_includes/selector.html
jarmak-nv: Hold-over from old logic, missed it when deleting old Docker-related logic.

_includes/selector.html
jarmak-nv: Resolved in https://github.com/rapidsai/docs/pull/414/commits/cb8e582dd6c332903e2cff27ed440967d7d0f222

_includes/selector.html
jarmak-nv: Resolved in https://github.com/rapidsai/docs/pull/414/commits/cb8e582dd6c332903e2cff27ed440967d7d0f222

install/install.md
bdice: The rule I go by is to always capitalize ARM (which really implies ARM SBSA for us, meaning “ARM in the server flavor and not Jetson which is called L4T”) or instead say aarch64 (the proper name of the arch used by conda — less common in our docs).

install/install.md
jarmak-nv: Is the install page able to access the site data like the selector? ie could I do:

`mamba ... rapids={{ site.data.releases.stable.version }} ...`

bdice: I don’t know. Please try it! Let’s find out.

jarmak-nv: We sure can!

bdice: Awesome. I wonder how many places we've missed the opportunity to do this in the past... (though we have some places where it makes more sense to hardcode if the guidance is temporary and applicable only to one specific release).

bdice:
> Oh one other question I had for the group: with `libmamba` [being a default installed solver now](https://www.anaconda.com/blog/new-release-anaconda-distribution-2023-07-miniconda-23-5-0-and-more), thoughts on adding
> 
> `--solver=libmamba` to all conda commands?

We've wanted to stay in line with conda defaults as much as possible. I am not sure that updated versions of conda are widespread enough to support this by default. However, I'm _very_ eager for it to become the default and recommend that all users update their conda versions...

jarmak-nv:
>We've wanted to stay in line with conda defaults as much as possible. I am not sure that updated versions of conda are widespread enough to recommend this to all users. However, I'm very eager for it to become the default and recommend that all users update their  conda versions...

Yeah this is a good point, conda installs do often feel like a "set and forget" thing. It's not beautiful, but we could do:

```
conda create -n rapids-23.06 -c rapidsai -c conda-forge -c nvidia rapids=23.06 python=3.10 cuda-version=11.8 --solver=libmamba || \
conda create -n rapids-23.06 -c rapidsai -c conda-forge -c nvidia rapids=23.06 python=3.10 cuda-version=11.8
```

Not strongly attached to this, but think it's worth bringing up just because mamba is so much faster and if people can use it, they should.

jarmak-nv:
💯 to recommending it, the challenge is when to assume it. If we include `--solver=libmamba` in the base command we provide from the selector, we're assuming users already have that solver available which is only true if you're on the newest conda install, or are on a late 2022 install + performed the steps to install the libmamba solver.

However, the error you get from conda is very clear if you try to use `--solver=libmamba` and you don't have it installed. I don't think it's too much of a blocker, but it isn't a perfectly clean "we should do this" imo.

(My weak pref is we do include it and add a troubleshooting step in the docs for when libmamba isn't installed.)

mmccarty:
Well yes, we also need to add / link to instructions on how to install it. I don't suppose we want yet another options in the selector.

bdice:
> (My weak pref is we do include it and add a troubleshooting step in the docs for when libmamba isn't installed.)

I like this proposal! Let’s add it “always” in our commands and then give clear troubleshooting instructions on how to fix it if their conda version is too old. Include a copy of the error message so it’s easy to find. Maybe add a note at the bottom of the conda selector like “We add `--solver=libmamba` for speeding up the environment solving process. This requires conda>=X.Y.Z. Alternatively, use mamba to install.”

jarmak-nv:
remove cuda 11.4/11.5

ajschmidt8:
image

these `/rapids/utils` scripts are no longer included/relevant in the new images, so this section on the https://deploy-preview-414--docs-rapids-ai.netlify.app/install page can be removed

jarmak-nv:
This update really cleans the selector up/simplifies it. One more potential simplification - is anyone opposed to removing `Source` as in installation method? All it does is provide links to the repos READMEs/Contrib guides. The install page covers this well enough IMO and we should remove it.

ajschmidt8:
> This update really cleans the selector up/simplifies it. One more potential simplification - is anyone opposed to removing `Source` as in installation method? All it does is provide links to the repos READMEs/Contrib guides. The install page covers this well enough IMO and we should remove it.

+1 to this.

I never really wanted to include it to begin with.

As you mentioned, with these Docker changes, our selector is now dramatically simplified. So I think we should keep trimming the fat to make it super easy for our end-users to get started.

ajschmidt8:
Oh, and another thing.

We can remove the `ARCH` selector entirely now too.

That selector was only used for our Docker images, but our Docker images are now multi-architectural. So there are no longer separate image tags for `arm` vs `x86`.

ajschmidt8:
> I thought about this too, we'll need to figure out a very good place to convey both of the things below without it though
> 
> * ARM != Jetson
> * CUDA 12 + ARM + Conda doesn't work yet
> 
> Maybe not big enough reasons to keep, but IMO people don't read things like the notes in the same way they do when they see something become disabled. Mostly concerned about an excess of slack/issues/stack posts about people trying jetson or CUDA 12 on ARM.

Gotcha. I'm in favor of keeping the selector as simple as possible, so hopefully we can come to a consensus on a better place for this information.

I also just noticed that the note outlined below is now identical for `base`/`notebooks` images. So we can probably remove that too.

image


additionally, we can replace the two `docker pull` and `docker run` commands with just one `docker run` command that has the `--pull always` flag. e.g.:

```sh
docker run --gpus all --pull always --rm -it \
    --shm-size=1g --ulimit memlock=-1 --ulimit stack=67108864 \
    rapidsai/base:23.06-cuda11.8-py3.10
```

This has been supported since the `20.10.0` release on 2020-12-08 ([src](https://docs.docker.com/engine/release-notes/20.10/#20100)).

jarmak-nv:
Pushed some potentially contentious changes in the name of simplification. Need to run for now, the perfect time for the push!

- Removes Arch
- Makes pip/conda lines have `Standard` and `Choose Specific Packages` with a new line that appears if you select the second option

On the notes for docker, I think they should stay since there's no other mention of what is installed. On the pull, command, happy to drop that, will do so when I come back.

ajschmidt8 (commented):
Can we remove the commented lines from `selector.html`?

If we need to retrieve those lines, we can always use the `git` history.

jarmak-nv:
> Can we remove the commented lines from selector.html?
>
> If we need to retrieve those lines, we can always use the git history.

Yeah this was always my plan, i just left them in for easy of retrieval if people felt strongly about keeping it in for signaling unsupported architectures, but might make a worse review experience so I'll delete them

bdice:
High level changes I'd like to see. I am going to review the code in a moment and might suggest changes there, to align with my observations in this comment.

- Clicking "Choose Specific Packages" should automatically select at least one package (cuDF) to avoid showing `pip install  --extra-index-url=https://pypi.nvidia.com`. It's a little awkward that selecting "no packages" is possible, but that'd be slightly nontrivial to enforce in the JavaScript code.
- Ordering of packages should be consistent between these two lists. Just flip the ordering of `cuSpatial` and `cuXfilter` for one of them.

image

- The second row of "Choose Specific Packages" says "PACKAGES" on the left side, again. This feels redundant. I think that title block for the second row can be empty.

image

- "TensorFlow requires CUDA 11.2." Do we know if this is really true? The package is built with 11.2 but conda succeeds at solving `conda create -n rapids-23.08 -c rapidsai-nightly -c conda-forge -c nvidia rapids=23.08 python=3.10 cuda-version=11.8 tensorflow --solver=libmamba` (note `cuda-version 11.8` instead of `11.2`) I'd believe any CUDA 11 might work. This is not important to solve for 23.08 but would be good to figure out and fix eventually.

install/install.md
```diff
@@ -47,16 +47,39 @@ Use the selector tool below to select your preferred method, packages, and envir
 ## Installation Troubleshooting
 
 ### **Conda Issues**
-<i class="fas fa-info-circle"></i> The dependency solver takes too long or never resolves: <br/>
-Update conda to use the new [libmamba solver](https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community/){: target="blank"} or use [Mamba directly](https://mamba.readthedocs.io/en/latest/installation.html){: target="_blank"}.
+<i class="fas fa-info-circle"></i> A `conda create error` occurs:<br/>
+RAPIDS has switched the default solver recommendation to [libmamba](https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community/){: target="_blank"}, a Mamba-powered Conda solver that is [now included with all Conda/Miniconda installations](https://www.anaconda.com/blog/new-release-anaconda-distribution-2023-07-miniconda-23-5-0-and-more){: target="_blank"} to significantly accelerate environment solving. If the error below occurs:
```
bdice, line 51:
```suggestion
RAPIDS has switched the default solver recommendation to [libmamba](https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community/){: target="_blank"}, a Mamba-powered Conda solver that is [now included with all Conda/Miniconda installations](https://www.anaconda.com/blog/new-release-anaconda-distribution-2023-07-miniconda-23-5-0-and-more){: target="_blank"} to significantly accelerate environment solving. The error output shows:
```

install/install.md
```diff
@@ -47,16 +47,39 @@ Use the selector tool below to select your preferred method, packages, and envir
...
+```
+mamba create -n rapids-pytorch-cu12 -c rapidsai -c pytorch-nightly -c conda-forge -c nvidia rapids={{ site.data.releases.stable.version }} cuda-version=12.0 pytorch pytorch-cuda=12.1
+```
 
 ### **Docker Issues**
-<i class="fas fa-info-circle"></i> Jupyter Lab is not accessible:<br/>
-If the server has not started or needs to be restarted / stop, use the included [start/stop script](#docker-startstop). Note this may change in the near future releases.
+<i class="fas fa-exclamation-triangle"></i> RAPIDS `23.08` brings significant Docker changes. <br/>
+To learn more about these changes, please see the [RAPIDS Container README](https://hub.docker.com/r/rapidsai/rapidsai){: target="_blank"}. Some key notes below:
+- `Development` images are no longer being published, RAPIDS now recommends [Dev Containers](https://code.visualstudio.com/docs/devcontainers/containers){: target="_blank"} for development
```
bdice, line 75: I don't think we can make a recommendation to use dev containers yet, since it's not available in the 23.08 release.

install/install.md
```diff
@@ -115,15 +138,26 @@ All provisioned systems need to be RAPIDS capable. Here's what is required:
...
 
  **Note**: RAPIDS is tested with and officially supports the versions listed above. Newer CUDA and driver versions may also work with RAPIDS. See [CUDA compatibility](https://docs.nvidia.com/deploy/cuda-compatibility/index.html) for details. 
+
+## **CUDA 12 Support**
+
+### **Docker and Conda**
+
+- <i class="fas fa-info-circle"></i> CUDA 12 packages and images currently support CUDA 12.0
+- <i class="fas fa-info-circle"></i> CUDA 11 packages and images can be used on a system with a CUDA 12 driver because they include their own CUDA toolkit
+- <i class="fas fa-info-circle"></i> ARM is not currently supported on CUDA 12 packages and images, use CUDA 11 or pip packages for ARM support
```
bdice, line 155: I think packages/images should be extra clear (conda packages / Docker images) so that people skimming this section don't confuse anything.
```suggestion
- <i class="fas fa-info-circle"></i> CUDA 12 conda packages and Docker images currently support CUDA 12.0
- <i class="fas fa-info-circle"></i> CUDA 11 conda packages and Docker images can be used on a system with a CUDA 12 driver because they include their own CUDA toolkit
- <i class="fas fa-info-circle"></i> ARM is not currently supported by CUDA 12 conda packages or Docker images, use CUDA 11 or pip packages for ARM support
```

install/install.md
```diff
@@ -210,9 +244,9 @@ docker run --gpus all nvcr.io/nvidia/k8s/cuda-sample:nbody nbody -gpu -benchmark
 **4b. Legacy Docker Users.** Docker CE v18 & [nvidia-docker2](https://github.com/NVIDIA/nvidia-docker/wiki/Installation-(version-2.0)){: target="_blank"} users will need to replace the following for compatibility:
 `docker run --gpus all` with `docker run --runtime=nvidia`
 
-<br/><br/>
+<br/>
 
-**JupyterLab.** Defaults will run [JupyterLab](https://jupyterlab.readthedocs.io/en/stable/){: target="_blank"} on your host machine at port: `8888`.
+### **JupyterLab.** Defaults will run [JupyterLab](https://jupyterlab.readthedocs.io/en/stable/){: target="_blank"} on your host machine at port: `8888`.
```
bdice, line 249: It's a little awkward to make a full sentence with links/code font into a header. It seems like "JupyterLab" is supposed to be treated like "Running Multi-Node..." below, and that symmetry was broken here.

What does "Defaults will run..." mean? Do we need to specify that Jupyter only runs for the "notebooks" container?

install/install.md
```diff
@@ -210,9 +244,9 @@ docker run --gpus all nvcr.io/nvidia/k8s/cuda-sample:nbody nbody -gpu -benchmark
 **4b. Legacy Docker Users.** Docker CE v18 & [nvidia-docker2](https://github.com/NVIDIA/nvidia-docker/wiki/Installation-(version-2.0)){: target="_blank"} users will need to replace the following for compatibility:
 `docker run --gpus all` with `docker run --runtime=nvidia`
 
-<br/><br/>
+<br/>
```
bdice, line 247: Question: do we need to tell users how to get into a bash prompt when using the new containers? Greg asked this when using the new containers (since they drop you into an ipython shell by default) and I think this is not obvious to a lot of folks who aren't container experts. I'm not sure what the old container default behavior was, but that'd be helpful to know if it used bash or something else when no entry point was provided. Is there a good place to include a bit of guidance like `docker run ... /bin/bash`?

install/install.md
```diff
@@ -239,12 +263,12 @@ bash /rapids/utils/stop-jupyter.sh
...
 
 ## **pip**
-RAPIDS 23.06 pip packages of cuDF, dask-cuDF, cuML, cuGraph, cuSpatial, RMM, and RAFT are available for CUDA 11 and CUDA 12 on the NVIDIA Python Package Index.
+RAPIDS pip packages are available for CUDA 11 and CUDA 12 on the NVIDIA Python Package Index.
 
 ### **pip Additional Prerequisites**
 <i class="fas fa-info-circle"></i> The CUDA toolkit version on your system must match the pip CUDA version you install (`-cu11` or `-cu12`). <br/>
 <i class="fas fa-info-circle"></i> **glibc version:** x86_64 wheels require glibc >= 2.17. <br/>
-<i class="fas fa-info-circle"></i> **glibc version:** ARM architecture (aarch64) wheels require glibc >= 2.31 (only ARM Server Base System Architecture is supported).
+<i class="fas fa-info-circle"></i> **glibc version:** ARM architecture (aarch64) wheels require glibc >= 2.32 (only ARM Server Base System Architecture is supported).
```
bdice, line 271: Is this true for CUDA 11 or only CUDA 12? (Or is this the glibc requirement from dask-sql?)

I'm surprised because Ubuntu 20.04 wouldn't even meet this criterion, because [Ubuntu 20.04 uses glibc 2.31](https://packages.ubuntu.com/focal/glibc-source). That'd be very confusing, because I think ARM + Ubuntu 20.04 works fine in practice... (I haven't tested this and would like clarification if anyone knows).

bdice: I did a bit more research and found that CUDA 12.0.1 supports glibc >=2.28 on ARM. https://docs.nvidia.com/cuda/archive/12.0.1/cuda-installation-guide-linux/index.html

This is my understanding from the conda-forge side of things too, I think we just need glibc >=2.28 for CUDA 12.

_includes/selector.html
```diff
@@ -9,6 +9,7 @@
     .selector-bg {
         background-color: #9943ff;
         padding: 1rem;
+        border-radius: 5px;
```
bdice, line 12: I'm sometimes a fan of rounded rectangles but I think this selector looked fine without it. No strong feelings; it's fine either way and I will avoid bikeshedding this further. :)

_includes/selector.html
```diff
@@ -307,13 +310,21 @@
...
-                        <template x-for="package in pip_packages">
+                        <template x-for="package in packages">
                             <div x-on:click="(e) => packagesClickHandler(e, package)"
                                 x-bind:class="{'active': active_packages.includes(package)}" class="option"
                                 x-html="getPackageHTML(package)">
                             </div>
                         </template>
                     </div>
+                    <div class="options-section" x-show="active_packages != 'Standard'">
+                        <div class="option-label">Packages</div>
```
bdice, line 321:
```suggestion
                        <div class="option-label"><!-- Second row of packages --></div>
```

_includes/selector.html
```diff
@@ -379,30 +373,25 @@
...
-            arches: ["amd", "arm"],
             img_loc: ["NGC", "Dockerhub"],
-            img_types: ["core", "standard"],
-            packages: ["Standard", "cuDF", "cuML", "cuGraph", "cuSpatial", "cuXFilter", "cuSignal", "cuCIM"],
-            pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial", "cuCIM"],
-            additional_packages: ["Dask SQL", "JupyterLab", "Plotly Dash", "Graphistry", "PyCaret", "TensorFlow", "Xarray-Spatial"], // , "RTXpy" will return later
-            img_options: ["notebooks", "development"],
+            img_types: ["Base", "Notebooks"],
+            packages: ["Standard", "Choose Specific Packages"],
+            additional_pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuXFilter", "cuSpatial", "cuCIM"],
```
bdice, line 392:
```suggestion
            additional_pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial", "cuXFilter", "cuCIM"],
```

_includes/selector.html
```diff
@@ -484,7 +460,7 @@
                     pkgs = ["rapids"];
                 }
 
-                var pkgs_vers = pkgs.map(pkg => this.highlightPkgOrImg(pkg) + "=" + rapids_version + " ").join("").toLowerCase();
+                var pkgs_vers = pkgs.filter(pkg => pkg !== "Choose Specific Packages").map(pkg => this.highlightPkgOrImg(pkg) + "=" + rapids_version + " ").join("").toLowerCase();
```
bdice, line 463: Maybe this can be a way of enforcing at least one package? The empty install command is awkward imo.
```suggestion
                if ((pkgs.length + this.active_additional_packages.length) == 0) {
                    return "Select at least one package.";
                }
                var pkgs_vers = pkgs.filter(pkg => pkg !== "Choose Specific Packages").map(pkg => this.highlightPkgOrImg(pkg) + "=" + rapids_version + " ").join("").toLowerCase();
```

_includes/selector.html
```diff
@@ -513,115 +489,84 @@
...
+
                 var libraryToPkg = (pkg) => {
                     pkg = pkg.toLowerCase();
                     if (pkg === "cucim") return pkg;
                     return pkg + cuda_suffix;
                 }
-                var pkgs = this.active_packages.map(libraryToPkg).join(" ");
-                var isArm = this.active_arch === "arm";
-                var cupy_inst = "";
+                if (this.active_packages[0] === 'Standard') {
```
bdice, line 498: Maybe this can be a way of enforcing at least one package? The empty install command is awkward imo.
```suggestion
                if (this.active_packages.length === 0) {
                    return "Select at least one package.";
                } else if (this.active_packages[0] === 'Standard') {
```

_includes/selector.html
```diff
@@ -513,115 +489,84 @@
...
             getImgLoc(loc) {
                 if (loc === "NGC") return 'NGC (Stable Only) <i class="fa-regular fa-share-nodes"></i>';
                 if (loc === "Dockerhub") return 'Docker Hub (Stable and Nightly) <i class="fa-docker fab"></i>';
                 return loc;
             },
             getDockerNotes() {
-                var core_pkgs = this.rapids_meta_pkgs;
+                if (this.active_cuda_ver.startsWith("12")) {
+                        var pkgs_html = this.rapids_meta_pkgs.filter(pkg => pkg !== "cuSignal").map(pkg => "<code>" + pkg + "</code>").join(", ");
+                        notes = [...notes, "RAPIDS CUDA 12.0 packages do not support ARM architectures"];
```
bdice, line 551:
```suggestion
                        notes = [...notes, "RAPIDS Docker images for CUDA 12 do not currently support ARM architectures"];
```

_includes/selector.html
```diff
@@ -513,115 +489,84 @@
...
                 return [this.note_prefix + " The selected image contains the following packages:<br>" + pkgs_html];
             },
             getCondaNotes() {
                 var notes = [];
                 notes = [...notes, "RAPIDS currently doesn't support <code>channel_priority: strict</code>; use <code>channel_priority: flexible</code> instead"];
                 if (this.active_packages.length === 1 && this.active_packages[0] === "Standard") {
-                    var pkgs_html = this.rapids_meta_pkgs.map(pkg => "<code>" + pkg + "</code>").join(", ");
-                    notes = [...notes, "The <code>Standard</code> selection contains the following packages:<br>" + pkgs_html];
+                    if (this.active_cuda_ver.startsWith("12")) {
+                        notes = [...notes, "RAPIDS CUDA 12.0 packages do not support ARM architectures"];
```
bdice, line 563:
```suggestion
                        notes = [...notes, "RAPIDS conda packages for CUDA 12 do not currently support ARM architectures"];
```

_includes/selector.html
```diff
@@ -728,6 +640,11 @@
                 if (this.active_release === "Nightly" && method === "pip") isDisabled = true;
                 return isDisabled;
             },
+            disableUnsupportedPackage(package) {
+                var isDisabled = false;
+                if (this.active_cuda_ver.startsWith("12") && this.active_method === "Conda" && package === "cuSignal") isDisabled = true;
```
bdice, line 645: Do we need to disable cuSignal for 23.10 nightlies? See https://docs.rapids.ai/notices/rsn0032/.
```suggestion
                if (package === "cuSignal" && ((this.active_cuda_ver.startsWith("12") && this.active_method === "Conda") || this.active_release === "Nightly")) isDisabled = true;
```

bdice (commented):
This review was more involved than I expected... 🙃 Most of these comments are fairly minor, and could be handled in later iterations if needed.

install/install.md
bdice: It'd be okay to say "coming soon." But probably not advisable to point to something that doesn't exist yet.

exactlyallan:
...In my designer days I didnt get why devs used CLIs so much. Now that Im on both sides I see how a UI means a factorial increase in complexity with each little added thing.

install/install.md
jarmak-nv: We actually didn't make Ubuntu 20.04 images for ARM in 23.06: https://github.com/rapidsai/docker/pull/551

The issue came up during notebook testing during release, maybe overlooked for a little while but with the overhaul notebook testing gets run as part of the build so hopefully shouldn't happen again.

_includes/selector.html
jarmak-nv: I don't think it looked _bad_ or anything, but was inconsistent with the rest of the page since all other buttons and code blocks have rounded corners and I like consistency and it was a quick change.

_includes/selector.html
jarmak-nv: Put in a different way of pre-setting cuDF if they go with `choose your own adventure` in https://github.com/rapidsai/docs/pull/414/commits/c73bfc773a76f0dc55aa253c9cd3c43b5699eec0. I try to stay away from errors whenever possible and I think the vast majority of users will have a mouse-click saved by pre-populating cuDF. 

Let me know what you think - happy to iterate more, IMO this is one of the major ways people get RAPIDS, and for new users it's their first experience with us, so it's worth the effort to get right.

install/install.md
jarmak-nv: Maybe at the top in the troubleshooting 23.08 docker overhaul highlights?

install/install.md
jarmak-nv: Updated in https://github.com/rapidsai/docs/pull/414/commits/3a31245cb4cd3305101b91430544afc1a06db4cd - hopefully helpful, also 10000% my assumption on what the intention was, I just thought it was better as a header and failed to hit enter to break it into a newline after **JupyterLab**

install/install.md
bdice: Okay, thanks for the clarification. That sounds right, it’s not just a CUDA requirement but also that bug.

bdice: (I mentioned dask-sql above, it’s the same bug here.)

_includes/selector.html
bdice: I think this should be a “yes and” solution.
- Yes, enable cudf by default like you did in c73bfc7
- Also show an error if no packages are selected (e.g. by clicking cudf off) by applying this suggesion

install/install.md
bdice: That’d be fine with me. I think that’s mostly important if the old containers behaved differently. Then you could consider putting it in the “major differences from old containers” section.

jarmak-nv: Added in https://github.com/rapidsai/docs/pull/414/commits/898912f4996959d4f61872c3fb3e18ab97f30066

install/install.md
```diff
@@ -47,16 +47,42 @@ Use the selector tool below to select your preferred method, packages, and envir
...
+The installation method below may allow RAPIDS CUDA 12.0 packages to coexist with PyTorch CUDA 12.1 nightly packages if there is a hard-requirement of CUDA 12 but it is currently unsupported:
+```
+mamba create -n rapids-pytorch-cu12 -c rapidsai -c pytorch-nightly -c conda-forge -c nvidia rapids={{ site.data.releases.stable.version }} cuda-version=12.0 pytorch pytorch-cuda=12.1
+```
 
 ### **Docker Issues**
-<i class="fas fa-info-circle"></i> Jupyter Lab is not accessible:<br/>
-If the server has not started or needs to be restarted / stop, use the included [start/stop script](#docker-startstop). Note this may change in the near future releases.
+<i class="fas fa-exclamation-triangle"></i> RAPIDS `23.08` brings significant Docker changes. <br/>
+To learn more about these changes, please see the [RAPIDS Container README](https://hub.docker.com/r/rapidsai/rapidsai){: target="_blank"}. Some key notes below:
```
raydouglass, line 74: Based on https://github.com/rapidsai/docker/pull/559:

```suggestion
To learn more about these changes, please see the [RAPIDS Container README](https://hub.docker.com/r/rapidsai/base){: target="_blank"}. Some key notes below:
```

install/install.md
```diff
@@ -47,16 +47,42 @@ Use the selector tool below to select your preferred method, packages, and envir
...
+- `Development` images are no longer being published, in the coming releases RAPIDS will roll out [Dev Containers](https://code.visualstudio.com/docs/devcontainers/containers){: target="_blank"} for development
+   - See cuSpatial for an example and information on [RAPIDS' usage of Dev Containers](https://github.com/rapidsai/cuspatial/tree/main/.devcontainer){: target="_blank"}
+- All images are Ubuntu-based
+   - CUDA 11.2 images are Ubuntu `20.04`
+   - All other images are Ubuntu `22.04`
+- All images are multiarch (x86_64 and ARM)
+   - CUDA 12 is not yet supported when using Docker images on ARM architecture
+- The `Base` image starts in an ipython shell
+   - To run bash commands inside the ipython shell prefix the command with `!`
+   - To run the image without the ipython shell add `/bin/bash` to the end of the `docker run` command
```
raydouglass, line 84: Should we link the RSN (https://docs.rapids.ai/notices/rsn0030/) or issue (https://github.com/rapidsai/docker/issues/539) that outlines all of the changes?

install/install.md
raydouglass: The `Breaking changes` section of https://github.com/rapidsai/docker/issues/539 addresses many situations to get the "old" functionality back.

_includes/selector.html
```diff
@@ -503,7 +491,7 @@
                     add_pkgs_ln
                 ].filter(Boolean).join(" \\\n");
 
-                return lines;
+                return lines + " --solver=libmamba";
```
raydouglass, line 494: I think the `--solver=libmamba` arg should go after `conda create` and before `-n rapids-23.08`.

_includes/selector.html
```diff
@@ -513,115 +501,87 @@
...
-                    imgVariant,
-                    (isDevel ? "dev" : ""),
+                    imgLoc + "rapidsai/" + imgVariant,
                     (isNightly ? "nightly" : ""),
-                    (isArm ? "arm64" : ""),
                 ].filter(Boolean).join("-");
 
                 var imgTag = [
-                    (isNightly ? this.getNightlyVersion() : this.getStableVersion()),
+                    (isNightly ? this.getNightlyVersion() + "a" : this.getStableVersion()),
```
raydouglass, line 543: I can't find exactly where this happens, but when it is a nightly docker version, the only change is adding the `a`, so the code needs updating so that it doesn't output `rapidsai/base-nightly:23.10a-cuda11.8-py3.10`. Note the `-nightly` being appended to the repo name.

jarmak-nv: Nice catch! Fixed in https://github.com/rapidsai/docs/pull/414/commits/7ac7dcd8fc00ae5038e1302dc00ed68c69c408bd

_includes/selector.html
jarmak-nv: Functionally it doesn't need to, but if there's a style/hierarchy we should adhere to I'm happy to change. I just like that it's at the end so if someone has a problem with it, they can delete it easily.

install/install.md
jarmak-nv: Yeah I'll link that issue, it's very thorough and easy to read (ty!)

install/install.md
jakirkham: The Dask-SQL GLIBC bug was recently fixed: https://github.com/dask-contrib/dask-sql/pull/1201

jarmak-nv:
I tested that all of the `extra packages` can resolve successfully with CUDA 12, didn't do any functionality testing, but it does solve!

_includes/selector.html
jarmak-nv: Updated to match the new rapids.ai ordering. https://github.com/rapidsai/rapids.ai/pull/328/commits/22e3bda7229a3f06d3df4a2c262dda73f717ff01

raydouglass (approved):

install/install.md
```diff
@@ -47,16 +47,43 @@ Use the selector tool below to select your preferred method, packages, and envir
...
+- Use the classic solver by removing `--solver=libmamba` from the `conda create` command provided by the selector
+
+<i class="fas fa-info-circle"></i> CUDA 12.0 ARM packages are not yet available:<br/>
+Conda-forge does not yet support the minimum required glibc (2.32) for CUDA 12 on ARM. For ARM support, please use CUDA 11.
+
+<i class="fas fa-info-circle"></i> At the time of writing, there is no stable CUDA 12 release of PyTorch: <br/>
+PyTorch currently only has nightly builds for CUDA 12.1, stable builds are limited to CUDA 11. <br/>
+The installation method below may allow RAPIDS CUDA 12.0 packages to coexist with PyTorch CUDA 12.1 nightly packages if there is a hard-requirement of CUDA 12 but it is currently unsupported:
+```
+mamba create -n rapids-pytorch-cu12 -c rapidsai -c pytorch-nightly -c conda-forge -c nvidia rapids={{ site.data.releases.stable.version }} cuda-version=12.0 pytorch pytorch-cuda=12.1
```
jakirkham, line 69: Do we want to use the `libmamba` solver here too?

```suggestion
conda create --solver=libmamba -n rapids-pytorch-cu12 -c rapidsai -c pytorch-nightly -c conda-forge -c nvidia rapids={{ site.data.releases.stable.version }} cuda-version=12.0 pytorch pytorch-cuda=12.1
```

jarmak-nv: I like standardization - committed!

ajschmidt8 (commented):
Quick question about the release selector logic behavior.

If you perform the following steps:

1. Go to selector https://deploy-preview-414--docs-rapids-ai.netlify.app/install#selector
2. Click `Docker` as the method
3. Click `Docker Hub` as the image location
4. Click `Nightly` as the release
5. Click `Stable` as the release

After the last step above, the selector re-selects `NGC` as the image location.

Is this intentional?

I would expect that changing the release should not change the image location since the `Stable` release is also on `Docker Hub`.

Is this an attempt by us to coerce people into using the NGC variants?

_includes/selector.html
```diff
@@ -513,115 +501,86 @@
...
+                } else {
+                        var pkgs_html = this.rapids_meta_pkgs.map(pkg => "<code>" + pkg + "</code>").join(", ");
+                    }
 
-                if (this.active_img_type === "standard") {
-                    core_pkgs = [...core_pkgs, "dask-sql"];
-                }
-                var pkgs_html = core_pkgs.map(pkg => "<code>" + pkg + "</code>").join(", ");
-                return [this.note_prefix + " The selected image contains the following packages:<br>" + pkgs_html];
+                return [this.note_prefix + cuda12 + " The selected image contains the following packages:<br>" + pkgs_html];
```
ajschmidt8, line 570: These changes might need to be tweaked.

They introduce some visual differences (see screenshots below) for the Docker notes compared to the other notes.

Namely, the line and icon spacing is different.

To fix this, each note should be its own entry in the returned array.


<details><summary>Click to Expand - Docker Notes</summary>
<p>

image

</p>
</details> 

<details><summary>Click to Expand - Conda Notes</summary>
<p>

image

</p>
</details>

_includes/selector.html
```diff
@@ -728,6 +654,11 @@
                 if (this.active_release === "Nightly" && method === "pip") isDisabled = true;
                 return isDisabled;
             },
+            disableUnsupportedPackage(package) {
+                var isDisabled = false;
+                if (package === "cuSignal" && ((this.active_cuda_ver.startsWith("12") && this.active_method === "Conda") || this.active_release === "Nightly")) isDisabled = true;
+                return isDisabled;
+            },
```
ajschmidt8, line 661: Is this logic working as expected?

I'm still able to select `cusignal` with conda, CUDA 12, and nightly selected.

image

jarmak-nv:
On the NGC stable image location swap, not really intended, just an artifact of the logic used here https://github.com/rapidsai/docs/blob/e2db3c90da5dcfbb113c5a91309d0004f5562d0d/_includes/selector.html#L668-L670

Easy enough to change and happy to do so (unless we do have an incentive to move people to the NGC image).

_includes/selector.html
jarmak-nv: oooo no, will debug - nice catch.

_includes/selector.html
jarmak-nv: Will make Docker notes match the Conda notes formatting

_includes/selector.html
jarmak-nv: Found the issue - when I added in the second line for specific package selection, I failed to add the disabled info to the options-container class.

Resolved in https://github.com/rapidsai/docs/pull/414/commits/a854dbe4cd82a8a7efd2e2c293a08082f89f84ac

_includes/selector.html
jarmak-nv: Took a few tries because the VSCode preview I use doesn't quite match the committed version, but should be resolved in https://github.com/rapidsai/docs/pull/414/commits/5427ebf4e35b4dea52c211704dbd4dcbff53a370

exactlyallan:
@jarmak-nv et all, this is looking good. Great tool we have here.

Very minor thing but in the css for the "command" text id make it match so the command label stays to the top:
```
.cmd-label {
    color: white;
    width: 6em;
    text-transform: uppercase;
    font-weight: 600;
    margin-top: 1rem;
    margin-right: 0.6em;
    vertical-align: top;
}
```

_includes/selector.html
ajschmidt8, line 1: looks like `cusignal` is still clickable even when it's disabled (see screenshot).

adding `if (this.isDisabled(e.target)) return;` as the first line of the click handlers should resolve this.

can you make sure all the click handlers include that line?


image

ajschmidt8 (commented):
almost there! there are a lot of changes in this release :slightly_smiling_face:

ajschmidt8:
> @jarmak-nv et all, this is looking good. Great tool we have here.
> 
> Very minor thing but in the css for the "command" text id make it match so the command label stays to the top:
> 
> ```
> .cmd-label {
>     color: white;
>     width: 6em;
>     text-transform: uppercase;
>     font-weight: 600;
>     margin-top: 1rem;
>     margin-right: 0.6em;
>     vertical-align: top;
> }
> ```


I think this should also include `text-align: right`.

It looks like the other labels have that property, but not the command label.

jakirkham (approved):
Thanks Ben! 🙏

Looks good from my perspective. Will defer to others on JS, etc.

ajschmidt8 (approved):
this is awesome.

the release selector is sooo much simpler as a result of the Docker image changes.

_includes/selector.html
```diff
@@ -389,9 +377,9 @@
             img_loc: ["NGC", "Dockerhub"],
             img_types: ["Base", "Notebooks"],
             packages: ["Standard", "Choose Specific Packages"],
-            additional_pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuXFilter", "cuSpatial", "cuCIM"],
+            additional_pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial", "cuXFilter", "cuCIM"],
```
bdice, line 380: I just noticed that we're inconsistent about the casing of `cuxfilter`, we have `cuXfilter` and `cuXFilter` written in this page. The README and docs say `cuxfilter` with no capital letters. @exactlyallan Can you verify if the name should be written `cuxfilter`?

_includes/selector.html
```diff
@@ -452,18 +440,22 @@
                 var pkgs = this.active_packages;
                 var py_cuda_pkgs = [this.highlightPkgOrImg("python") + "=" + python_version, this.highlightPkgOrImg("cuda-version") + "=" + cuda_version].join(" ");
                 var conda_channels = [rapids_channel, "conda-forge", "nvidia"]
-                    .map(ch => this.highlightFlag("-c") + " " + ch + " ")
+                    .map(ch => "-" + this.highlightFlag("c") + " " + ch + " ")
```
bdice, line 443: I think the dashes are usually highlighted the same way as the flag name?
```suggestion
                    .map(ch => this.highlightFlag("-c") + " " + ch + " ")
```

My vim highlighting looks like:

image

_includes/selector.html
```diff
@@ -452,18 +440,22 @@
...
+
                 if (this.active_packages.length === 1 && this.active_packages[0] === "Standard") {
                     pkgs = ["rapids"];
                 }
 
                 var pkgs_vers = pkgs.filter(pkg => pkg !== "Choose Specific Packages").map(pkg => this.highlightPkgOrImg(pkg) + "=" + rapids_version + " ").join("").toLowerCase();
                 var all_pkgs = pkgs_vers + py_cuda_pkgs;
 
-                var conda_create_ln = this.highlightCmd('conda') + " create " + this.highlightFlag("-n") + " rapids-" + rapids_version + " " + conda_channels;
+                var conda_create_ln = this.highlightCmd('conda') + " create --" + this.highlightFlag("solver") + "=libmamba " + this.highlightFlag("-n") + " rapids-" + rapids_version + " " + conda_channels;
```
bdice, line 458:
```suggestion
                var conda_create_ln = this.highlightCmd('conda') + " create " + this.highlightFlag("--solver") + "=libmamba " + this.highlightFlag("-n") + " rapids-" + rapids_version + " " + conda_channels;
```

_includes/selector.html
```diff
@@ -479,12 +471,13 @@
...
                 ].filter(Boolean).join(" \\\n");
 
-                return lines + " --solver=libmamba";
+                return lines;
             },
             getpipCmdHtml() {
-                var pip_install = "pip install ";
-                var index_url = " --extra-index-url=https://pypi.nvidia.com";
+                var pip_install = `${this.highlightCmd("pip")} install`;
+                var index_url = `--${this.highlightFlag("extra-index-url")}=https://pypi.nvidia.com`;
```
bdice, line 478:
```suggestion
                var index_url = `${this.highlightFlag("--extra-index-url")}=https://pypi.nvidia.com`;
```

_includes/selector.html
```diff
@@ -534,7 +530,7 @@
                 var fullImg = (this.highlightPkgOrImg(imgRepo) + ":" + imgTag).toLowerCase();
 
                 var indent = "    ";
-                var cmd = this.highlightCmd("docker") + " run --" + this.highlightFlag("gpus") + " all --pull always --" + this.highlightFlag("rm") + " -" + this.highlightFlag("it") + " \\\n" +
+                var cmd = this.highlightCmd("docker") + " run --" + this.highlightFlag("gpus") + " all --" + this.highlightFlag("pull") + " always --" + this.highlightFlag("rm") + " -" + this.highlightFlag("it") + " \\\n" +
```
bdice, line 533:
```suggestion
                var cmd = this.highlightCmd("docker") + " run " + this.highlightFlag("--gpus") + " all " + this.highlightFlag("--pull") + " always " + this.highlightFlag("--rm") + " " + this.highlightFlag("-it") + " \\\n" +
```

_includes/selector.html
```diff
@@ -534,7 +530,7 @@
...
                     indent + "--" + this.highlightFlag("shm-size") + "=1g --" + this.highlightFlag("ulimit") + " memlock=-1 --" + this.highlightFlag("ulimit") + " stack=67108864" + " \\\n" +
```
bdice, line 534:
```suggestion
                    indent + this.highlightFlag("--shm-size") + "=1g " + this.highlightFlag("--ulimit") + " memlock=-1 " + this.highlightFlag("--ulimit") + " stack=67108864" + " \\\n" +
```

_includes/selector.html
```diff
@@ -582,7 +582,7 @@
             getpipNotes() {
                 var notes = [];
                 notes = [...notes, "<code>cuDF</code>, <code>dask-cuDF</code>, <code>cuML</code>, <code>cuGraph</code>, <code>cuSpatial</code>, and <code>cuXfilter</code> pip packages are hosted on the NVIDIA Python Package Index<br>",
-                         'pip installation supports Python <code>3.9</code>, and <code>3.10</code><br>'];
+                    'pip installation supports Python <code>3.9</code>, and <code>3.10</code><br>'];
```
bdice, line 585:
```suggestion
                    'pip installation supports Python <code>3.9</code> and <code>3.10</code><br>'];
```

bdice (approved):
A few remaining comments, but otherwise LGTM!

_includes/selector.html
bdice: I like `cuxfilter`.

_includes/selector.html
ajschmidt8: interesting. okay. I will update in a bit. I just made them match the existing highlights.

_includes/selector.html
jarmak-nv: Lower-cased all instances in https://github.com/rapidsai/docs/pull/414/commits/6323860c0192398c60bbd9b5381def1d1290112d

ajschmidt8:
> Somewhat perplexed why the deploy preview isn't showing the cuProj updates. My local preview shows fine, but it's like they're not there at all in netlify.

Must've been a hiccup on Netlify's end. I reran the build from our admin dashboard and it looks like it's showing now.

jakirkham:
Is there anything else needed here?

jarmak-nv:
@jakirkham It's all good - @ajschmidt8 just wants a moment to verify/modify highlighting per https://github.com/rapidsai/docs/pull/414#discussion_r1281168769 but otherwise no concerns.

jakirkham:
Maybe let's do final reviews. Think we can punt on smaller stylistic changes

bdice (approved):

jakirkham (approved):

jakirkham:
Thanks all! 🙏

PR #428, opened 2023-09-05
Update RSN Notices 19,21, and 28

caryr35:
Updating RSN Notices 19,21, and 28 to completed


_notices/rsn0021.md
```diff
@@ -9,7 +9,7 @@ notice_id: 21 # should match notice number
 notice_pin: true # set to true to pin to notice page
```
raydouglass, line 9: I think we should unpin these notices. And probably many others.

raydouglass (approved):

aravenel (approved):

PR #435, opened 2023-10-04
Pascal GPU deprecation release v23.12

caryr35:
RAPIDS team is planning to deprecate Pascal GPUs support in its `v23.12` release and remove support in its `v24.02` release


_notices/rsn0034.md
```diff
@@ -0,0 +1,36 @@
...
+notice_topic: Platform Support Change
+notice_rapids_version: "v23.12"
+notice_created: 2023-10-04
+# 'notice_updated' should match 'notice_created' until an update is made
+notice_updated: 2023-10-04
+---
+
+## Overview
+
+RAPIDS team is planning to deprecate Pascal GPUs support in its `v23.12` release and remove support in its `v24.02` release
```
aravenel, line 29:
```suggestion
RAPIDS team is planning to deprecate support for Pascal GPUs in its `v23.12` release and to remove support in its `v24.02` release. `v23.12` will be the last version of RAPIDS to support Pascal GPUs.
```

_notices/rsn0034.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Impact
+
+ RAPIDS will no longer support Pascal GPU's
```
aravenel, line 33:
```suggestion
 RAPIDS will no longer support Pascal GPUs effective with its `v24.02` release. Users of Pascal GPUs will be able to continue to use RAPIDS `v23.12`.  
```

_notices/rsn0034.md
bdice: Should we mention something like “Compute Capability 7.0 or higher will be required”?

aravenel: Good idea. I think that fits better in the "impact" section below, so have added there.

PR #450, opened 2023-11-15
Selector updates: pip, RAFT and pytorch

jarmak-nv:
This PR does 5 main things:
1. Adds pip nightly versions :tada:
   1. cucim now uses the `-cu11/12` name in line with the rest of RAPIDS
3. Adds RAFT as a selectable RAPIDS package, and notes its inclusion in the meta package
4. Adds pytorch to the selector
5. Changes the default CUDA to `12.0`
6. Removes all mention of cuSignal as `23.08` was the final release prior to cuPy ingesting its functionality

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -460,6 +460,12 @@
                     pkgs.push("cuSpatial");
                     pkgs.push("cuProj");
                 }
+                if (pkgs.includes("RAFT")) {
+                    pkgs = pkgs.filter(pkg => pkg !== "RAFT");
+                    pkgs.push("pylibraft");
+                    pkgs.push("raft-dask");
+                }
+
 
                 var pkgs_vers = pkgs.filter(pkg => pkg !== "Choose Specific Packages").map(pkg => this.highlightPkgOrImg(pkg) + "=" + rapids_version + " ").join("").toLowerCase();
                 var all_pkgs = pkgs_vers + py_cuda_pkgs;
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -505,7 +505,7 @@
                 }
                 else {
                     index_url = `--${this.highlightFlag("extra-index-url")}=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple`;
-                    cuda_suffix = cuda_suffix + ">={{ site.data.releases.nightly.version }}.0a0,<{{ site.data.releases.nightly.version }}\"";
+                    cuda_suffix = cuda_suffix + ">={{ site.data.releases.nightly.version }}.0a0,<={{ site.data.releases.nightly.version }}\"";
                     var libraryToPkg = (pkg) => {
                         pkg = pkg.toLowerCase();
                         if (pkg == "cuspatial/cuproj") return ["\"cuspatial" + cuda_suffix, "\"cuproj" + cuda_suffix];
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -359,7 +359,7 @@
         Alpine.data('rapids_selector', () => ({
             // default values
             active_python_ver: "3.10",
-            active_cuda_ver: "11.8",
+            active_cuda_ver: "12.0",
             active_pip_cuda_ver: "12.0",
             active_method: "Conda",
             active_release: "Stable",
@@ -377,11 +377,11 @@
             img_loc: ["NGC", "Dockerhub"],
             img_types: ["Base", "Notebooks"],
             packages: ["Standard", "Choose Specific Packages"],
-            additional_pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuCIM"],
-            additional_rapids_packages: ["cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuSignal", "cuCIM"],
-            additional_packages: ["Dask-SQL", "JupyterLab", "Plotly Dash", "Graphistry", "TensorFlow", "Xarray-Spatial"],
+            additional_pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuCIM", "RAFT"],
+            additional_rapids_packages: ["cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuCIM", "RAFT"],
+            additional_packages: ["Dask-SQL", "JupyterLab", "Plotly Dash", "Graphistry", "TensorFlow", "Xarray-Spatial", "Pytorch"],
             note_prefix: "<i class='fas fa-info-circle text-blue'></i>",
-            rapids_meta_pkgs: ["cuDF", "cuML", "cuGraph", "cuSpatial", "cuProj", "cuxfilter", "cuSignal", "cuCIM"],
+            rapids_meta_pkgs: ["cuDF", "cuML", "cuGraph", "cuSpatial", "cuProj", "cuxfilter", "cuCIM", "RAFT"],
             getStableVersion() {
                 return "{{ site.data.releases.stable.version }}";
             },
@@ -481,7 +481,12 @@
             },
             getpipCmdHtml() {
                 var pip_install = `${this.highlightCmd("pip")} install`;
-                var index_url = `--${this.highlightFlag("extra-index-url")}=https://pypi.nvidia.com`;
+                if (this.active_release === "Stable") {
+                    index_url = `--${this.highlightFlag("extra-index-url")}=https://pypi.nvidia.com`;
+                }
+                else {
+                    index_url = `--${this.highlightFlag("extra-index-url")}=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple`;
+                }
                 var cuda_suffix = "-cu12";
                 var indentation = "    ";
 
@@ -491,7 +496,7 @@
 
                 var libraryToPkg = (pkg) => {
                     pkg = pkg.toLowerCase();
-                    if (pkg === "cucim") return pkg;
+                    if (pkg === "cucim" && this.active_release === "Stable") return pkg; // Remove this line in 23.12 when stable cucim out on our index
                     if (pkg == "cuspatial/cuproj") return "cuspatial" + cuda_suffix + " cuproj" + cuda_suffix;
                     return pkg + cuda_suffix;
                 }
@@ -501,7 +506,6 @@
                     var pkgs = this.additional_pip_packages.map(libraryToPkg);
                 } else {
                     var pkgs = this.active_packages
-                        .filter(pkg => pkg !== "Choose Specific Packages" && pkg !== "cuSignal")
                         .map(libraryToPkg);
                 }
 
@@ -551,7 +555,7 @@
             getDockerNotes() {
                 var notes = [];
                 if (this.active_cuda_ver.startsWith("12") && this.active_release === "Stable") {
-                    var pkgs_html = this.rapids_meta_pkgs.filter(pkg => pkg !== "cuSignal").map(pkg => "<code>" + pkg + "</code>").join(", ");
+                    var pkgs_html = this.rapids_meta_pkgs.map(pkg => "<code>" + pkg + "</code>").join(", ");
                     var cuda12 = "Only nightly RAPIDS Docker images for CUDA 12 currently support ARM architectures<br>";
                     notes = [...notes, cuda12]
                 } else {
@@ -568,7 +572,7 @@
                 if (this.active_packages.length === 1 && this.active_packages[0] === "Standard") {
                     if (this.active_cuda_ver.startsWith("12") && this.active_release === "Stable") {
                         notes = [...notes, "Only nightly RAPIDS conda packages for CUDA 12 currently support ARM architectures"];
-                        var pkgs_html = this.rapids_meta_pkgs.filter(pkg => pkg !== "cuSignal").map(pkg => "<code>" + pkg + "</code>").join(", ");
+                        var pkgs_html = this.rapids_meta_pkgs.map(pkg => "<code>" + pkg + "</code>").join(", ");
                     } else {
                         var pkgs_html = this.rapids_meta_pkgs.map(pkg => "<code>" + pkg + "</code>").join(", ");
                     }
@@ -622,7 +626,6 @@
             disableUnsupportedRelease(release) {
                 var isDisabled = false;
                 if (this.active_img_loc === "NGC" && this.active_method === "Docker" && release === "Nightly") isDisabled = true;
-                if (this.active_method === "pip" && release === "Nightly") isDisabled = true;
                 return isDisabled;
             },
             disableUnsupportedCuda(cuda_version) {
@@ -645,12 +648,10 @@
             },
             disableUnsupportedMethod(method) {
                 var isDisabled = false;
-                if (this.active_release === "Nightly" && method === "pip") isDisabled = true;
                 return isDisabled;
             },
             disableUnsupportedPackage(package) {
                 var isDisabled = false;
-                if (package === "cuSignal" && ((this.active_cuda_ver.startsWith("12") && this.active_method === "Conda") || this.active_release === "Nightly")) isDisabled = true;
                 return isDisabled;
             },
             isDisabled(e) {
@@ -666,9 +667,6 @@
             },
             cudaClickHandler(e, version) {
                 if (this.isDisabled(e.target)) return;
-                if (version === "12.0") {
-                    this.active_packages = this.active_packages.filter(pkg => pkg !== "cuSignal");
-                }
                 this.active_cuda_ver = version;
             },
             pipCUDAClickHandler(e, version) {
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -508,7 +508,7 @@
                     cuda_suffix = cuda_suffix + ">={{ site.data.releases.nightly.version }}.0a0\"";
                     var libraryToPkg = (pkg) => {
                         pkg = pkg.toLowerCase();
-                        if (pkg == "cuspatial/cuproj") return ["cuspatial" + cuda_suffix, "cuproj" + cuda_suffix];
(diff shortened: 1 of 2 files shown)


raydouglass:
@jarmak-nv Looks like the command is weird for pip choose specific packages:
image

jarmak-nv:
Ah yes, good call - I was too aggressive when deleting filters for cuSignal. Should be resolved.

I do have an outstanding question though: after doing some testing with the nightly install command using the anaconda index, I need to use the `--pre` flag for pip to resolve correctly (throws an error otherwise). 

Is there a way we can not use that flag, or should we expect to need it because of dask-cuda anyway?

It's not currently in the selector, so I will add it in if there's no other approach (my weak pref would be we only select stable when possible, but this is the nightly so maybe prereleases are preferable anyway)

charlesbluca:
Out of interest, do you have the conflicts that came up without `--pre`? Wonder if there's a way to specify pre-releases for only the packages causing conflicts rather than enabling the flag globally

jarmak-nv:
Here's the output from running without `--pre` 

<details><summary>Click to expand</summary>

```
(anaconda-test-nopre) (base) picard@Enterprise-G:~/anaconda-test-nopre$ pip install cudf-cu12 --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple
Looking in indexes: https://pypi.org/simple, https://pypi.anaconda.org/rapidsai-wheels-nightly/simple
Collecting cudf-cu12
  Using cached cudf-cu12-23.10.1.tar.gz (6.8 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error

  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [17 lines of output]
      Traceback (most recent call last):
        File "/home/picard/anaconda-test-nopre/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/home/picard/anaconda-test-nopre/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
        File "/home/picard/anaconda-test-nopre/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
        File "/tmp/pip-build-env-g9ae53ys/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 355, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
        File "/tmp/pip-build-env-g9ae53ys/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 325, in _get_build_requires
          self.run_setup()
        File "/tmp/pip-build-env-g9ae53ys/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 507, in run_setup
          super(_BuildMetaLegacyBackend, self).run_setup(setup_script=setup_script)
        File "/tmp/pip-build-env-g9ae53ys/overlay/lib/python3.10/site-packages/setuptools/build_meta.py", line 341, in run_setup
          exec(code, locals())
        File "<string>", line 81, in <module>
      RuntimeError: Bad params
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
``` 
</details>

The same error occurs if I attempt things like:
`pip install cudf-cu12>=23.12.* ...`

Digging into [pip docs](https://pip.pypa.io/en/stable/cli/pip_install/#install-pre), it looks like if we define the exact pre-release version it may be able find it, but otherwise we need the global --pre. Even if we had access to the exact builds, it would make some ugly commands (but I guess we're copy/pasting, so not 100% unreasonable)

Confirmed by doing `pip install cudf-cu12==23.12.0a699 ...` successfully without the `--pre` flag.

bdice:
> Out of interest, do you have the conflicts that came up without `--pre`? Wonder if there's a way to specify pre-releases for only the packages causing conflicts rather than enabling the flag globally

These worked for me:
- `pip install "cudf-cu12>=0.0.0a0" --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple`
- `pip install "cudf-cu12>=23.12.0a0" --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple`

I tried following https://github.com/rapidsai/cudf/blob/8deb3dd7573000e7d87f18a9e2bbe39cf2932e10/ci/build_wheel.sh#L28-L35 and attempted `cudf-cu12==23.12,>=0.0.0a0` but that didn't work for me with pip 23.1.2.

jarmak-nv:
> * `pip install "cudf-cu12>=0.0.0a0" --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple

Interesting! This worked for me, but without the quotes (below), it does not. Is this new/intended behavior in pip? I don't see it documented in the [pip docs](https://pip.pypa.io/en/stable/cli/pip_install/#install-pre), but I might be missing it!

bdice:
> My thinking is that CUDA 12 has been out for almost a year now and we should prefer it.

Yes. New users will be getting CUDA 12. It seems like most users reporting issues are on CUDA 12 now. We should be defaulting to CUDA 12 for all installation methods.

bdice:
> Interesting! This worked for me, but without the quotes (below), it does not.

`bash` will eat that `>` and pipe the output to a file named `=0.0.0a0` or something like that. Check your working directory and you might see it. :wink: If you're writing a constraint for pip (or conda!) on the command line, you always want to wrap it with quotes unless it's only using equals signs.

jarmak-nv:
> `bash` will eat that `>` and pipe the output to a file named `=0.0.0a0` or something like that.

I'm one of the lucky 10,000!
xkcd image

Okay, so consensus time:

would we rather have the selector output something like:
```
pip install \
    --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple \
    "cudf-cu12>=23.12.0a0" "dask-cudf-cu12>=23.12.0a0" "cuml-cu12>=23.12.0a0" "cugraph-cu12>=23.12.0a0"
    and so on
```

or
```
pip install \
    --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple --pre \
    cudf-cu12 dask-cudf-cu12 cuml-cu12 cugraph-cu12
    and so on
```

I'm lightly in favor of `--pre` because it's a much easier to read install command, but it's only weakly in favor and some day if we decide to make a meta-package on pip it wouldn't be so unwieldy.

bdice:
The `--pre` will have additional impacts, like fetching pre-release versions of all dependencies: numpy, numba, etc. It's less predictable and harder to test because of upstream changes that can break RAPIDS. I vote to stick to `"cudf-cu12>=23.12.0a0"` for safety, but I'd want to get a second opinion from @vyasr on this since he's scoped this more than I have.

vyasr:
> Is there a way we can not use that flag, or should we expect to need it because of dask-cuda anyway?
> 
> It's not currently in the selector, so I will add it in if there's no other approach (my weak pref would be we only select stable when possible, but this is the nightly so maybe prereleases are preferable anyway)

The current behavior is expected. Our nightlies are versioned as pre-releases (the `...a*`) and pip will not download those by default when using a vanilla `pip install $pkg` command. Previously we were versioning our nightlies differently to avoid this requirement by adding an extra release segment like `23.12.00.00000001`, but we decided to move away from that because that causes fairly problematic behavior if a user installs a pre-release then wants to upgrade to an official release because `23.12.00.00000001 > 23.12.00.00`. Since the existing alpha versioning logic is something pip users are already accustomed to, we decided that moving to normal alphas was the right decision before making our nightlies publicly available. 

You have two options to make it allow pre-releases, 1) adding `--pre` to the command, or 2) changing the version specifier for the request to one that allows for nightlies. The second case would be (as you've already found) doing something like `pip install cudf-cu11==23.12.00a23` to request a specific alpha or doing `pip install cudf-cu11>=23.12.00a0` and allowing it to choose the latest version with that lower bound, but including alphas (by having `a0` at the end of that version specifier). 

> The --pre will have additional impacts, like fetching pre-release versions of all dependencies: numpy, numba, etc. It's less predictable and harder to test because of upstream changes that can break RAPIDS. I vote to stick to "cudf-cu12>=23.12.0a0" for safety (despite it being very verbose), but I'd want to get a second opinion from @vyasr on this since he's scoped this more than I have.

Bradley is spot on here and it's why I've been recommending the `>=23.12.0a0` approach. We're opening our users up to more errors (and therefore ourselves up to more bug reports) if we implicitly install a numpy alpha that breaks behaviors that we rely on internally.

_includes/selector.html
```diff
@@ -377,11 +377,11 @@
             img_loc: ["NGC", "Dockerhub"],
             img_types: ["Base", "Notebooks"],
             packages: ["Standard", "Choose Specific Packages"],
-            additional_pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuCIM"],
-            additional_rapids_packages: ["cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuSignal", "cuCIM"],
-            additional_packages: ["Dask-SQL", "JupyterLab", "Plotly Dash", "Graphistry", "TensorFlow", "Xarray-Spatial"],
+            additional_pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuCIM", "RAFT"],
+            additional_rapids_packages: ["cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuCIM", "RAFT"],
+            additional_packages: ["Dask-SQL", "JupyterLab", "Plotly Dash", "Graphistry", "TensorFlow", "Xarray-Spatial", "Pytorch"],
```
bdice, line 382:
```suggestion
            additional_packages: ["Dask-SQL", "JupyterLab", "Plotly Dash", "Graphistry", "TensorFlow", "Xarray-Spatial", "PyTorch"],
```

_includes/selector.html
```diff
@@ -501,8 +506,8 @@
                     var pkgs = this.additional_pip_packages.map(libraryToPkg);
                 } else {
                     var pkgs = this.active_packages
-                        .filter(pkg => pkg !== "Choose Specific Packages" && pkg !== "cuSignal")
-                        .map(libraryToPkg);
+                    .filter(pkg => pkg !== "Choose Specific Packages")
+                    .map(libraryToPkg);
```
bdice, line 510: The indent indicates line continuation. I'd keep it.
```suggestion
                        .filter(pkg => pkg !== "Choose Specific Packages")
                        .map(libraryToPkg);
```

_includes/selector.html
```diff
@@ -481,34 +481,54 @@
...
                 }
 
-                var libraryToPkg = (pkg) => {
-                    pkg = pkg.toLowerCase();
-                    if (pkg === "cucim") return pkg;
-                    if (pkg == "cuspatial/cuproj") return "cuspatial" + cuda_suffix + " cuproj" + cuda_suffix;
-                    return pkg + cuda_suffix;
+                // Change index depending on stable vs nightly for pip
+                // Also add versioning commands for nightly installs to prevent --pre usage
+                // This has dupilicate code, but makes for easier edits in the future
```
bdice, line 493:
```suggestion
                // This has duplicate code, but makes for easier edits in the future
```

_includes/selector.html
```diff
@@ -481,34 +481,54 @@
...
                     cuda_suffix = "-cu11";
                 }
 
-                var libraryToPkg = (pkg) => {
-                    pkg = pkg.toLowerCase();
-                    if (pkg === "cucim") return pkg;
-                    if (pkg == "cuspatial/cuproj") return "cuspatial" + cuda_suffix + " cuproj" + cuda_suffix;
-                    return pkg + cuda_suffix;
+                // Change index depending on stable vs nightly for pip
+                // Also add versioning commands for nightly installs to prevent --pre usage
```
bdice, line 492: Is this what you mean?
```suggestion
                // Also add versioning commands for nightly installs so that --pre is unnecessary
```

_includes/selector.html
```diff
@@ -481,34 +481,54 @@
...
+                        .filter(pkg => pkg !== "Choose Specific Packages")
                         .map(libraryToPkg);
                 }
 
                 if (pkgs.length === 1 && pkgs[0] === "cucim") {
                     index_url = "";
                 }
 
+                //for every 3rd package add a new line with a "\" character
+                for (var i = 3; i < pkgs.length; i += 4) {
```
bdice, line 528: This behavior change should match the comment. Currently the second (and later) line of packages will contain 4 packages instead of 3. It gets a little too long.
```suggestion
                for (var i = 3; i < pkgs.length; i += 3) {
```

_includes/selector.html
```diff
@@ -481,34 +481,54 @@
...
+                        return "\"" + pkg + cuda_suffix;
+                    }
                 }
+
                 if (this.active_packages.length === 1 && this.active_packages[0] === "Choose Specific Packages") {
                     return "Select at least one package.";
                 } else if (this.active_packages[0] === 'Standard') {
                     var pkgs = this.additional_pip_packages.map(libraryToPkg);
                 } else {
                     var pkgs = this.active_packages
```
bdice, line 518: Can we sort the list of active packages whenever it is changed, so that the list of active packages matches the display order? (cuDF, dask-cuDF, cuML, ...)

Otherwise the list of packages that shows up appears to be dependent on the order in which you select the boxes.

jarmak-nv: hey sorry yeah I should have converted this to draft - I just needed to push it to see how it rendered sorry!

jarmak-nv: Done in https://github.com/rapidsai/docs/pull/450/commits/1c9802fed82bdb2af540513bb89fd0c2f1672978

jarmak-nv:
I just did a test run of this, and realized that during burndown-release doing `>=23.12.0a0` isn't enough; if you run that right now you get `24.02` builds.

I think to be certain we need:
`"$package-cu12>=23.12.0a0,<23.12"` (I've tested this and it works on my end)

Does that seem right @vyasr @bdice?

_includes/selector.html
```diff
@@ -437,13 +437,16 @@
...
                 var cuda_version = this.active_cuda_ver;
-                var pkgs = this.active_packages;
                 var py_cuda_pkgs = [this.highlightPkgOrImg("python") + "=" + python_version, this.highlightPkgOrImg("cuda-version") + "=" + cuda_version].join(" ");
                 var conda_channels = [rapids_channel, "conda-forge", "nvidia"]
                     .map(ch => "-" + this.highlightFlag("c") + " " + ch + " ")
                     .join("");
                 var indentation = "    ";
 
+                // sort active_packages to appear in the same order as the additional_rapids_packages list
+                this.active_packages.sort((a, b) => this.additional_rapids_packages.indexOf(a) - this.additional_rapids_packages.indexOf(b));
```
bdice, line 447: Very clever! I had been brainstorming all the ways to implement this and I like your solution a lot. 👍

_includes/selector.html
```diff
@@ -481,37 +490,73 @@
...
+                        pkg = pkg.toLowerCase();
+                        if (pkg === "cucim") return pkg; // Remove this line in 23.12 when stable cucim is out on our index
+                        if (pkg === "cuspatial/cuproj") return ["cuspatial" + cuda_suffix, "cuproj" + cuda_suffix];
+                        if (pkg === "raft") return ["pylibraft" + cuda_suffix, "raft-dask" + cuda_suffix];
+                        return [pkg + cuda_suffix];
+                    }
+                }
+                else {
+                    index_url = `--${this.highlightFlag("extra-index-url")}=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple`;
+                    cuda_suffix = cuda_suffix + ">={{ site.data.releases.nightly.version }}.0a0,<={{ site.data.releases.nightly.version }}\"";
```
bdice, line 515: Should `<=` be `<`, to prevent stable package installation?
```suggestion
                    cuda_suffix = cuda_suffix + ">={{ site.data.releases.nightly.version }}.0a0,<{{ site.data.releases.nightly.version }}\"";
```

bdice (approved):

_includes/selector.html
jarmak-nv: I thought so too, but it actually won't resolve at all if you do that (I just double checked to make sure). 

We could do
`...,<23.12.0a$sufficiently-large-number`

Not sure what's less ideal, the $sufficiently-large-number, or having users get a stable 23.12 between 24.02's release and the site update (I think that's the only time it could happen.) It's already a long pip install command so not much lost in changing it...

bdice: Oof. I'd prefer `<23.12.0rc0` over `<23.12.0a9999`. But it's probably fine to permit stable builds. Can you revert 68a5cc2? Let's just roll with the previous `<=`.

bdice: cc: @vyasr if you have last minute thoughts here.

bdice: > having users get a stable 23.12 between 24.02's release and the site update

The user would also have to mix between the stable and nightly index -- I'm not worried about this happening in practice.

bdice: Maybe we can revert 68a5cc2, merge this, and then revisit it in a new PR for any later improvements. I'd like to make some changes to this page but don't want to introduce conflicts with this PR.

jarmak-nv: Done. Happy to merge so you can get changes in. Will do so in a couple hours just in case there are last minute comments from CCed people/you. Or if you want to get working on other changes asap feel free to merge when you're ready to work.

vyasr: Just seeing this post-merge, but FWIW I'm also not overly concerned about mixing nightlies and stable indexes, so I don't think we need to overengineer the command specifically to avoid the stable release. Also the whole point of the versioning changes was that nightlies are strictly ordered less than stable releases, so if the wrong package gets installed users should be able to use normal pip commands to upgrade without needing to go through manual uninstall/install cycles.

PR #467, opened 2024-01-17
Display UCXX in docs.rapids.ai/api

AyodeAwe:
UCXX docs is being deployed [here](https://github.com/rapidsai/ucxx/pull/164/files#diff-17e8ca0c6c748293e3ef7d9837679b7d09ffecb5307af0e85290cfbec2fb9d77R28) to s3 not with its own versioning convention (`0.36`, etc) but with RAPIDS versioning convention (`24.xx`, etc). This PR updates the publishing template script [here](https://github.com/rapidsai/docs/blob/main/_includes/api-docs.html) to overwrite the RAPIDS versions for UCXX to display the correct UCXX versions.

The changes in this PR should be sufficient to have `libucxx` docs display over at docs.rapids.ai/api

NOTE: the `cllink` anchor for `libucxx` is hidden as we currently do not have a CHANGELOG.md for the library. Adding one would imply significantly altering certain sections of our workflow which might not be worth the effort seeing that there is the intention to move `UCXX` to the [openucx](https://github.com/openucx) org.

### DIFF
--- a/_data/docs.yml
+++ b/_data/docs.yml
@@ -227,6 +227,17 @@ libs:
       legacy: 1
       stable: 1
       nightly: 1
+  libucxx:
+    name: libucxx
+    path: libucxx
+    desc: "UCXX is an object-oriented C++ interface for UCX, with native support for Python bindings."
+    ghlink: https://github.com/rapidsai/ucxx
+    cllink: https://github.com/rapidsai/ucxx/blob/main/CHANGELOG.md
+    versions:
+      # enable or disable links; 0 = disabled, 1 = enabled
+      legacy: 0
+      stable: 0
+      nightly: 1
   rapids-cmake:
     name: rapids-cmake
     path: rapids-cmake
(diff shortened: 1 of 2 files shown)


ajschmidt8 (commented):
This was merged prematurely.

I don't think we should have this much `ucxx` specific logic in the HTML templates.

The HTML template logic should be generic and leverage configuration information from `_data/docs.yml`.

Additionally, we shouldn't hardcode UCX version numbers in the HTML templates. This isn't maintainable.

They should be listed alongside the RAPIDS version numbers in `_data/releases.json` and referenced dynamically in the templates like the RAPIDS version numbers are.

Otherwise there is no single source of truth and we have to update version numbers in multiple areas during a release.

AyodeAwe:
Will open a PR to address this ^^

AyodeAwe:
Fixed in https://github.com/rapidsai/docs/pull/484

PR #479, opened 2024-02-16
Remove leading zeros from pip install commands

raydouglass:
I'm unsure when this changed, but `pip install` commands with leading zeros (not PEP-440 compliant) fail.

For example:
```
$ pip install --extra-index-url=https://pypi.nvidia.com cugraph-cu12==24.02.*
Looking in indexes: https://pypi.org/simple, https://pypi.nvidia.com
ERROR: Could not find a version that satisfies the requirement cugraph-cu12==24.02.* (from versions: 23.6.0, 23.6.1, 23.6.2, 23.8.0, 23.10.0, 23.12.0, 24.2.0)
ERROR: No matching distribution found for cugraph-cu12==24.02.*

$ pip install --extra-index-url=https://pypi.nvidia.com cugraph-cu12==24.2.*
Looking in indexes: https://pypi.org/simple, https://pypi.nvidia.com
Collecting cugraph-cu12==24.2.*
  Downloading https://pypi.nvidia.com/cugraph-cu12/cugraph_cu12-24.2.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (1311.2 MB)
```

So this PR removes the leading zeros.

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -488,6 +488,9 @@
 
                 return lines;
             },
+            removeLeadingZeros(version) {
+                return version.split(".").map(Number).join(".");
+            },
             getpipCmdHtml() {
                 var pip_install = `${this.highlightCmd("pip")} install`;
                 var cuda_suffix = "-cu12";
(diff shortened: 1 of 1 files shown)


bdice (approved):
Seems fine to me.

jakirkham:
Yeah this is a good idea

Think the tooling [should normalize them]( https://packaging.python.org/en/latest/specifications/version-specifiers/#integer-normalization ). If that is not happening, likely there is a bug somewhere

Regardless dropping the 0s ourselves is cleaner and should work the same

vyasr:
Agree that this ought to have worked already but no harm in us pre-normalizing.

PR #481, opened 2024-02-20
Adds RSN 38 for 24.02 MG cugraph `degree()` bug workaround.

rlratzel:
Adds RSN 38 for 24.02 MG cugraph `degree()` bug workaround.

### DIFF
--- a/_notices/rsn0038.md
+++ b/_notices/rsn0038.md
@@ -1,0 +1,52 @@
+---
+layout: notice
+parent: RAPIDS Support Notices
+grand_parent: RAPIDS Notices
+nav_exclude: true
+notice_type: rsn
+# Update meta-data for notice
+notice_id: 38 # should match notice number
+notice_pin: false # set to true to pin to notice page
+title: "Support for 24.02 cugraph Users"
+notice_author: RAPIDS Ops
+notice_status: In Progress
+notice_status_color: yellow
+# 'notice_status' and 'notice_status_color' combinations:
+#   "Proposal" - "blue"
+#   "Completed" - "green"
+#   "Review" - "purple"
+#   "In Progress" - "yellow"
+#   "Closed" - "red"
+notice_topic: Platform Support Change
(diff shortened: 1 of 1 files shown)


_notices/rsn0038.md
```diff
@@ -0,0 +1,52 @@
...
+result_df["degree"] //= 2
+```
+
+## Impact
+
+Impact is limited to RAPIDS 24.02 `cugraph` users using the graph `degree` functions (`G.degree()`, `G.in_degree()`, `G.out_degree()`) on undirected, multi-GPU distributed graphs.
+
+## Background
+
+A bug was noticed late in the 24.02 release process for `cugraph` users using the `degrees` functions (`G.degree()`, `G.in_degree()`, `G.out_degree()`) on undirected multi-GPU distributed graphs.  The bug causes the returned values to be doubled, and users of these functions for these graph types must apply a workaround to divide the result by two in order to obtain the correct values as seen in the example above.
```
bdice, line 50:
```suggestion
A bug was noticed late in the 24.02 release process for `cugraph` users using `degree` functions (`G.degree()`, `G.in_degree()`, `G.out_degree()`) on undirected multi-GPU distributed graphs.  The bug causes the returned values to be doubled, and users of these functions for these graph types must apply a workaround to divide the result by two in order to obtain the correct values as seen in the example above.
```

bdice (changes requested):
This generally feels like the kind of thing that would warrant a hotfix rather than a notice. Users of cuGraph probably do not look at RSNs for how to handle numerical bugs. RSNs are generally reserved for changes in platform support or package/dependency incompatibility.

Are we sure we don't want to hotfix this, rather than issue an RSN?

_notices/rsn0038.md
rlratzel: Thanks! I fixed this in the latest commit.

rlratzel:
Closing.  We'll backport the fix which addresses the issue described in this RSN to PB 2.1

PR #491, opened 2024-03-03
add note on putting channel_priority: flexible in your .condarc

raybellwaves:
Just adding a small note that users should put channel_priority: flexible in their .condarc. Apologies my linter made some other changes. Happy to revert if need be.

Hope this doesn't bloat the selector and still looks ok. I'm happy for it be moved elsewhere or left out if preferred.

Edit: Looks as expected in preview

Screenshot 2024-03-03 at 2 54 27 PM

Edit: if I give this a second thought you can also recommend using `--no-channel-priority` in the off chance a user needs strict channel priority in another project.

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -343,7 +343,8 @@
             <div class="cmd">
                 <div class="cmd-label">Command</div>
                 <div class="cmd-box">
-                    <pre class="highlight"><code><span x-ref="cmd" x-html="getCmdHtml" x-on:mouseup.once="copyToGA"></span></code></pre>
+                    <pre
+                        class="highlight"><code><span x-ref="cmd" x-html="getCmdHtml" x-on:mouseup.once="copyToGA"></span></code></pre>
                 </div>
             </div>
         </div>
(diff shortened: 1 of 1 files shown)


mmccarty:
Thanks @raybellwaves ! 
cc @jakirkham @bdice @msarahan

jakirkham:
Yeah we should note this somewhere

@jarmak-nv do you have thoughts on how we note this requirement?

jarmak-nv:
Agree this should be noted; I'm always hesitant to add things in the notes section of the selector though (where I would love to go is that only temporary/new items that significantly impact users in a new way go there.)

I think we should include this in the conda environment section instead, but 100% agree it's a good add. Thanks @raybellwaves!

jarmak-nv:
Hey @raybellwaves following up here - did you have plans on adding the note to the conda environment section, or should we close this PR and move it to an issue for followup?

raybellwaves:
I'll close this as i've opened https://github.com/rapidsai/docs/issues/501

PR #495, opened 2024-03-07
Remove dependency management page.

bdice:
This PR removes `depmgmt.md`. Its contents were mostly out of date and much of the content is redundant with external resources.


jarmak-nv (approved):
I was noodling on if we should just remove it vs your previous revision and I agree with this decision.

cc @raydouglass in case he has any concerns.

bdice:
@raydouglass proposed the full removal -- so we should be all good. I'll merge this.

Issue #498, opened 2024-03-21
[FEA] Add other install methods to Selector tool?


raybellwaves:
**Is your feature request related to a problem? Please describe.**

Wonder if it's worth adding other install methods into the selector tool (https://docs.rapids.ai/install#selector) such as uv (https://github.com/astral-sh/uv) and pixi (https://pixi.sh/latest/)?

**Describe the solution you'd like**
Perhaps additional boxes in the selector tool or comments around replacements e.g. `uv pip` for `pip`.

**Describe alternatives you've considered**
I understand adding a few new package management tools who release often runs the risk that it may break things. You also need additional documentation on using them and troubleshooting. They also may not be mature enough that's they are worth including.

**Additional context**
Recently saw ibis has pixi in their install docs https://ibis-project.org/install 

Recently got a nice speed up (275%) using `uv pip` vs `pip` on google colab (https://colab.research.google.com/drive/1d3_gb_4y2Xgvl6xBBdknfwUV8Heq-nzJ)

pixi may offer speed up downloading conda packages but I haven't worked a clean rapids install yet (https://github.com/prefix-dev/pixi/discussions/1029)

bdice:
I left some thoughts on the PR: https://github.com/rapidsai/docs/pull/510#issuecomment-2065170581

Overall, RAPIDS aims to support the most stable and well-recognized installation methods like `pip` and `conda`. If we see increased interest in using other tools, we can definitely re-evaluate the RAPIDS documentation in the future. For now, I'll close this issue since I don't think there are short-term actions. However, we can continue conversation and I'll reference this issue in the future if we see interest grow as tools become more mainstream. We welcome issues reports from users of `uv`, `pixi`, etc. if there are ways that RAPIDS packages need to be improved.

PR #510, opened 2024-04-14
add uv pip install method

raybellwaves:
Closes https://github.com/rapidsai/docs/issues/498

See the build at https://deploy-preview-510--docs-rapids-ai.netlify.app/install

Works as expected

Screenshot 2024-04-13 at 10 45 03 PM

Apologies for the linting edits as well

While uv pip offers a ~2-3 x RAPIDS install (4 min 19 s -> 1 min 34 s: https://colab.research.google.com/drive/1d3_gb_4y2Xgvl6xBBdknfwUV8Heq-nzJ) compared to pip it may not be as stable. However, from my experience the astral folks are pretty quick at fixing bugs (e.g. https://github.com/astral-sh/uv/issues/2477). I don't know if the docs here also need to include a section on how to download uv and troubleshoot.

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -343,7 +343,8 @@
             <div class="cmd">
                 <div class="cmd-label">Command</div>
                 <div class="cmd-box">
-                    <pre class="highlight"><code><span x-ref="cmd" x-html="getCmdHtml" x-on:mouseup.once="copyToGA"></span></code></pre>
+                    <pre
+                        class="highlight"><code><span x-ref="cmd" x-html="getCmdHtml" x-on:mouseup.once="copyToGA"></span></code></pre>
                 </div>
             </div>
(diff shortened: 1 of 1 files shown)


bdice:
@raybellwaves I'm really eager to play more with `uv` on my own system, but I don't think it's stable enough to be used for the default installation of RAPIDS. For the same reasons, we have stuck with `conda` as our default installation command even when `mamba` presented a compelling alternative (until `conda` added the `libmamba` solver, which is now the best of both worlds -- hopefully `pip` can borrow some ideas for improved performance from `uv`). I encourage users to try out `uv` but I don't think we're ready to make this switch.

I'll close this PR for cleanup -- but feel free to keep using `uv` and share your findings if you run into any issues with RAPIDS and `uv`.

PR #517, opened 2024-06-07
Remove librmm docs

raydouglass:
`librmm` docs were merged into `rmm` docs and `24.02` was the last version to have separated `librmm` docs.

Now that the legacy version is `24.04`, the `librmm` docs can be removed.

Ref: https://github.com/rapidsai/docs/pull/478
https://github.com/rapidsai/docs/actions/runs/9421083274/job/25954311807

### DIFF
--- a/_data/docs.yml
+++ b/_data/docs.yml
@@ -216,17 +216,6 @@ libs:
       legacy: 1
       stable: 1
       nightly: 1
-  librmm:
-    name: librmm
-    path: librmm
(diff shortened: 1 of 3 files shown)


bdice (changes requested):
We need a solution for libcudf's docs that currently depend on librmm's Doxygen docs.

See https://github.com/rapidsai/cudf/blob/39c5b86645dc61bf0c59d7bf733ca13872b46a44/cpp/doxygen/Doxyfile#L2229

ajschmidt8 (approved):

raydouglass:
> We need a solution for libcudf's docs that currently depend on librmm's Doxygen docs.

After some internal discussions, this does need to be addressed, but we don't want to block updating to the `24.06` docs, so going to override and merge this PR.

bdice:
Following up: https://github.com/rapidsai/rmm/pull/1604 re-adds librmm docs. From discussion with @raydouglass, we don't need to link to them publicly on docs.rapids.ai, as they are already included in the rmm docs (both C++ and Python). The librmm docs will just be used to cross-reference from libcudf docs.

PR #518, opened 2024-06-07
Fix ucxx docs & enable docs for newer libs

raydouglass:
Fixes pulling `libucxx` docs by adding the `ucxx_version` to `releases.json`.

Also enables various doc versions for newer libraries that now have legacy or stable docs.

I manually ran `./ci/download_from_s3.sh` locally to verify all of the docs exist.


bdice (approved):
Nit: maybe swap lines 9/10 so the fields are ordered as “version, ucxx_version, date.”

PR #521, opened 2024-07-17
Remove dask-sql from install selector

charlesbluca:


_includes/selector.html
```diff
@@ -427,8 +427,7 @@
             getAdditionalPkgName(pkg) {
                 var dask_prefix = this.active_release === "Nightly" ? "dask/label/dev::" : "";
```
bdice, line 428: @charlesbluca We have a variable `dask_prefix` in a couple places that appears to be unused now. Can we delete that?

jakirkham: Pushed a change to Charles' PR to do that: https://github.com/rapidsai/docs/pull/521/commits/d1da50494ef95daec0c82547f18434dca074a545

bdice (approved):

PR #524, opened 2024-07-22
Use CUDA version ranges in release selector.

bdice:
This PR uses CUDA version ranges in the release selector. This follows from discussions in https://github.com/rapidsai/docs/pull/522.

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -244,11 +244,11 @@
                         </div>
                     </template>
                 </div>
-                <div class="options-section" x-show="active_method !== 'pip'">
+                <div class="options-section" x-show="active_method === 'Conda'">
                     <div class="option-label">ENV. CUDA</div>
-                    <template x-for="version in cuda_vers">
-                        <div x-on:click="(e) => cudaClickHandler(e, version)"
-                            x-bind:class="{'active': version === active_cuda_ver, 'disabled': disableUnsupportedCuda(version)}"
+                    <template x-for="version in conda_cuda_vers">
+                        <div x-on:click="(e) => condaCUDAClickHandler(e, version)"
+                            x-bind:class="{'active': version === active_conda_cuda_ver, 'disabled': disableUnsupportedCuda(version)}"
                             class="option" x-text="'CUDA ' + version"></div>
                     </template>
                 </div>
@@ -258,6 +258,14 @@
                         <div x-on:click="(e) => pipCUDAClickHandler(e, version)"
                             x-bind:class="{'active': version === active_pip_cuda_ver}" class="option"
                             x-text="'CUDA ' + version"></div>
+                    </template>
+                </div>
+                <div class="options-section" x-show="active_method === 'Docker'">
+                    <div class="option-label">Image CUDA</div>
+                    <template x-for="version in docker_cuda_vers">
+                        <div x-on:click="(e) => dockerCUDAClickHandler(e, version)"
+                            x-bind:class="{'active': version === active_docker_cuda_ver, 'disabled': disableUnsupportedCuda(version)}"
+                            class="option" x-text="'CUDA ' + version"></div>
                     </template>
                 </div>
                 <div class="options-section" x-show="active_method !== 'pip'">
@@ -360,7 +368,7 @@
         Alpine.data('rapids_selector', () => ({
             // default values
             active_python_ver: "3.11",
-            active_cuda_ver: "12.2",
+            active_cuda_ver: "12.0 - 12.2",
             active_pip_cuda_ver: "12",
             active_method: "Conda",
             active_release: "Stable",
@@ -371,8 +379,9 @@
 
             // all possible values
             python_vers: ["3.9", "3.10", "3.11"],
-            cuda_vers: ["11.2", "11.8", "12.0", "12.2", "12.5"],
-            pip_cuda_vers: ["11.2 - 11.8", "12"],
+            conda_cuda_vers: ["11.4 - 11.8", "12.0 - 12.2", "12.0 - 12.5"],
+            pip_cuda_vers: ["11.4 - 11.8", "12"],
+            docker_cuda_vers: ["11.8", "12.0", "12.2", "12.5"],
             methods: ["Conda", "pip", "Docker"],
             releases: ["Stable", "Nightly"],
             img_loc: ["NGC", "Dockerhub"],
@@ -673,9 +682,9 @@
             },
             disableUnsupportedCuda(cuda_version) {
                 var isDisabled = false;
-                if (this.active_additional_packages.includes("TensorFlow") && (cuda_version !== "12.0")) isDisabled = true;
-                if (this.active_method === "Docker" && cuda_version < "11.8") isDisabled = true;
-                if (this.active_release === "Stable" && cuda_version === "12.5") isDisabled = true;
+                if (this.active_additional_packages.includes("TensorFlow") && (!cuda_version.startsWith("12")) isDisabled = true;
+                // Remove once stable releases support CUDA 12.5
+                if (this.active_release === "Stable" && cuda_version.includes("12.5")) isDisabled = true;
                 return isDisabled;
             },
             disableUnsupportedPython(python_version) {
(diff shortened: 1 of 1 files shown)


bdice:
@jarmak-nv @jakirkham Can you review and play with the site preview to see if this works like you expect?

I tried to get one button for conda to have CUDA 12.0-12.2 that would change to 12.0-12.5 when enabling nightlies but I didn't want to mess with any more of this selector logic at the moment.

_includes/selector.html
```diff
@@ -789,7 +808,7 @@
                     return;
                 }
                 this.active_additional_packages = [...this.active_additional_packages, package];
-                if (this.active_additional_packages.includes("TensorFlow") && (this.active_cuda_ver !== "12.0")) this.active_cuda_ver = "12.0";
+                if (this.active_additional_packages.includes("TensorFlow") && (!this.active_conda_cuda_ver.includes("12.0"))) this.active_conda_cuda_ver = "12.0 - 12.2";
```
jakirkham, line 811:
```suggestion
                if (this.active_additional_packages.includes("TensorFlow") && (!this.active_conda_cuda_ver.includes("12.0"))) this.active_conda_cuda_ver = this.conda_cuda_ver_stable;
```

jakirkham (commented):
Thanks Bradley! 🙏

Wonder if we can generalize this a bit to simplify the updating cycle for new CUDA versions

Took a first pass at this below, but likely more needs to change (like Docker). Also there may be better ways to capture the same logic (like a `bool` variable that simply states whether nightly and stable versions match)

jarmak-nv:
Logic seems good, no issues after lots of clicks/testing.

I think having both 12.0-12.2 and 12.0-12.5 buttons is a bit awkward though. While I like the specificity of ranges, I think I'm still lightly in favor of just `cuda 11` `cuda 12` and `cuda 12 nightly` or something.

jakirkham:
If we don't show the version ranges on the buttons, where do we specify these ranges?

jakirkham:
Thanks Bradley! 🙏

This is looking pretty good

<hr>

Noticed one thing with `pip` that I wanted to highlight (though maybe you already saw this)

Screenshot 2024-07-22 at 7 57 18 PM

All the others look ok. So probably something particular to the `pip` case

bdice:
@jakirkham Glad this is going in the right direction -- I'm still working on some improvements, and I'm using CI as my test for website rendering. I may push some commits that don't work. I will let you know when this is ready for a final look.

bdice:
@jakirkham @jarmak-nv I think this is ready to review.

_includes/selector.html
```diff
@@ -424,6 +434,25 @@
...
+                var cuda_version_info = {
+                    "Stable": {
+                        "11": ["11.4", "11.8"],
+                        "12": ["12.0", "12.2"]
+                    },
+                    "Nightly": {
+                        "11": ["11.4", "11.8"],
+                        "12": ["12.0", "12.5"]
+                    }
+                };
```
jakirkham, line 447: Wonder if we can collect this info next to the "all possible values" section. That way there would be one place to change all of the versions

_includes/selector.html
```diff
@@ -705,22 +733,26 @@
             releaseClickHandler(e, release) {
                 if (this.isDisabled(e.target)) return;
                 this.active_release = release;
-                if ( this.active_release === "Stable" && this.active_cuda_ver === "12.5") {
-                    this.active_cuda_ver = "12.2";
+                if (this.active_release === "Stable" && this.active_docker_cuda_ver === "12.5") {
+                    this.active_docker_cuda_ver = "12.2";
                 }
```
jakirkham, line 738: What do you think about capturing `12.5` & `12.2` in variables in the "all possible values" section?

Perhaps we could do something like...?

```suggestion
                if (this.active_release === "Stable" && this.active_docker_cuda_ver === this.nightly_docker_cuda_ver) {
                    this.active_docker_cuda_ver = this.stable_docker_cuda_ver;
                }
```

_includes/selector.html
```diff
@@ -673,8 +702,7 @@
             },
             disableUnsupportedCuda(cuda_version) {
                 var isDisabled = false;
-                if (this.active_additional_packages.includes("TensorFlow") && (cuda_version !== "12.0")) isDisabled = true;
-                if (this.active_method === "Docker" && cuda_version < "11.8") isDisabled = true;
+                if (this.active_additional_packages.includes("TensorFlow") && (!cuda_version.startsWith("12"))) isDisabled = true;
                 if (this.active_release === "Stable" && cuda_version === "12.5") isDisabled = true;
```
jakirkham, line 706: Curious if we could use a variable for `12.5` here as well

_includes/selector.html
```diff
@@ -789,7 +821,7 @@
                     return;
                 }
                 this.active_additional_packages = [...this.active_additional_packages, package];
-                if (this.active_additional_packages.includes("TensorFlow") && (this.active_cuda_ver !== "12.0")) this.active_cuda_ver = "12.0";
+                if (this.active_additional_packages.includes("TensorFlow") && this.active_conda_cuda_ver !== "12") this.active_conda_cuda_ver = "12";
```
jakirkham, line 824: Liking the simplification here :)

jakirkham (commented):
Thanks Bradley! 🙏

Played around with the install selector. Liked the ergonomics of these improvements. Also felt like it was clear from a user perspective what they were getting. Not to mention they could further refine it if needed (like picking a specific CUDA version or working with other dependencies that have such constraints). Tried various combinations of selections to make sure the output was looking as expected

Think it is great that the code is being generalized so that we have fewer hard-coded versions in it. That will only benefit us longer term

Noted a couple spots that still have some hard-coding (likely not a focus on the recent revisions). Am curious if there are ways we can turn those into variables as well. Would be interested to hear your thoughts on this 🙂

_includes/selector.html
bdice: I struggled with this. I don't know why, but I couldn't get the JavaScript to work happily when I did that. See my attempt here (I force-pushed it out of the history for this PR since it wasn't going well): https://github.com/rapidsai/docs/commit/3a6e9a595313af4c1c9b5eb6f816f9c25cc222a0

jarmak-nv:
> If we don't show the version ranges on the buttons, where do we specify these ranges?

Yeah that's the challenge, my thought is that the selector should not be the main way of communicating what ranges of CUDA 12 the user can install alongside RAPIDS, but it does serve that purpose. Ranges work for this and the current implementation looks very nice/clean so I'm happy here.

---

I like how this works and the back-end changes! Thanks @bdice 

Only concern is we no longer highlight that the nightly has a higher CUDA version available than the stable. Not sure if there's a straightforward path to resolve this or if it's a big deal. 

---

Also on testing I noticed something I think might be considered undesirable, when selecting `pip` as the installation method, it moves you to CUDA 12 regardless of your previous selection. I think recently we've been settling on "don't change something the user has set unless the selection cannot work" and in that design theme we'd want to remove this behavior.

I think its controlled by the code here:
https://github.com/rapidsai/docs/blob/ab1bbce41968893d9f0d03aff60fa06fac5725f7/_includes/selector.html#L771C1-L773C18

bdice:
> Only concern is we no longer highlight that the nightly has a higher CUDA version available than the stable. Not sure if there's a straightforward path to resolve this or if it's a big deal.

I am leaning towards "this is not a big deal." I think most users are happy as long as some CUDA minor version for the selected major version works with their environment.

bdice:
> Also on testing I noticed something I think might be considered undesirable, when selecting pip as the installation method, it moves you to CUDA 12 regardless of your previous selection.

I will revisit this in a follow-up PR. I have a solution in mind but I think this PR is in a fairly complete state and would be better to merge as is.

_includes/selector.html
bdice: Could we leave this for a later PR? I think this ties in with the [above thread](https://github.com/rapidsai/docs/pull/524#discussion_r1688249043), where I struggled to get a "structured" JSON format to work.

_includes/selector.html
bdice: Could we leave this for a later PR? I think this ties in with the [above thread](https://github.com/rapidsai/docs/pull/524#discussion_r1688249043), where I struggled to get a "structured" JSON format to work.

jakirkham:
> > Only concern is we no longer highlight that the nightly has a higher CUDA version available than the stable. Not sure if there's a straightforward path to resolve this or if it's a big deal.
> 
> I am leaning towards "this is not a big deal." I think most users are happy as long as some CUDA minor version for the selected major version works with their environment.

Agreed. The window where these differ is usually short. So don't think it is worth the effort

jakirkham:
FWIW am ok merging and iterating in follow ups. This feels like a substantial improvement

jarmak-nv (approved):
> FWIW am ok merging and iterating in follow ups. This feels like a substantial improvement

Same! Thanks @bdice

jakirkham:
Thank you both! 🙏

PR #534, opened 2024-08-27
install selector: Drop python 3.9 for nightlies, default to CUDA 12.5 for stable and nightly

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/88.

Proposes the following changes to the install selector:

* disabling Python 3.9 for nightlies
* defaulting to CUDA 12.5 for stable and nightlies
* dropping CUDA 12.2 from the Docker selector

### How I tested this

Followed the instructions in the README. Ran locally and saw what I expected... Python 3.9 disabled when I tried selected `Nightly`.

Screenshot 2024-08-27 at 1 45 02 PM

But enabled when I selected `Stable`

Screenshot 2024-08-27 at 1 46 06 PM

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ docker run --rm \
   --volume="$PWD:/srv/jekyll" \
   --publish [::1]:4000:4000 \
   jekyll/jekyll \
-  jekyll serve
+  bash -c 'rm -rf ./_site && jekyll serve'
 ```
 
 ## PR submissions
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -370,7 +370,7 @@
             active_python_ver: "3.11",
             active_conda_cuda_ver: "12",
             active_pip_cuda_ver: "12",
-            active_docker_cuda_ver: "12.2",
+            active_docker_cuda_ver: "12.5",
             active_method: "Conda",
             active_release: "Stable",
             active_img_type: "Base",
@@ -379,7 +379,9 @@
             active_additional_packages: [],
 
             // all possible values
-            python_vers: ["3.10", "3.11"],
+            python_vers: ["3.9", "3.10", "3.11"],
+            python_vers_stable: ["3.9", "3.10", "3.11"],
+            python_vers_nightly: ["3.10", "3.11"],
             conda_cuda_vers: ["11", "12"],
             pip_cuda_vers: ["11.4 - 11.8", "12"],
             docker_cuda_vers: ["11.8", "12.0", "12.2", "12.5"],
@@ -708,6 +710,11 @@
             },
             disableUnsupportedPython(python_version) {
                 var isDisabled = false;
+                if (this.active_release === "Stable") {
+                    isDisabled = !this.python_vers_stable.includes(python_version)
+                } else if (this.active_release === "Nightly") {
+                    isDisabled = !this.python_vers_nightly.includes(python_version)
+                }
                 return isDisabled;
             },
             disableUnsupportedImgType(type) {
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -384,7 +384,7 @@
             python_vers_nightly: ["3.10", "3.11"],
             conda_cuda_vers: ["11", "12"],
             pip_cuda_vers: ["11.4 - 11.8", "12"],
-            docker_cuda_vers: ["11.8", "12.0", "12.2", "12.5"],
+            docker_cuda_vers: ["11.8", "12.0", "12.5"],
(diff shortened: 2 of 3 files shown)


_includes/selector.html
```diff
@@ -380,9 +380,11 @@
 
             // all possible values
             python_vers: ["3.9", "3.10", "3.11"],
+            python_vers_stable: ["3.9", "3.10", "3.11"],
+            python_vers_nightly: ["3.10", "3.11"],
             conda_cuda_vers: ["11", "12"],
             pip_cuda_vers: ["11.4 - 11.8", "12"],
-            docker_cuda_vers: ["11.8", "12.0", "12.2", "12.5"],
+            docker_cuda_vers: ["11.8", "12.0", "12.5"],
```
jameslamb, line 387: 12.5 images have been available since 24.08: https://github.com/rapidsai/docker/pull/689

12.2 images will be dropped in 24.10: https://github.com/rapidsai/docker/pull/696

CONTRIBUTING.md
```diff
@@ -22,7 +22,7 @@ docker run --rm \
   --volume="$PWD:/srv/jekyll" \
   --publish [::1]:4000:4000 \
   jekyll/jekyll \
-  jekyll serve
+  bash -c 'rm -rf ./_site && jekyll serve'
```
jameslamb, line 25: Following these testing steps as documented worked... but only the first time.

This command leaves behind a `_site/` directory with file permissions showing the default user inside the container as the owner. Where that user differs from the default user on the system where you run `docker run`, the second attempt to run this fails like this:

```text
chown: _site/assets/data: Permission denied
chown: _site/assets/data: Permission denied
chown: _site/assets: Permission denied
chown: _site/assets: Permission denied
chown: _site: Permission denied
chown: _site: Permission denied
```

Removing that directory at startup every time avoids that tiny bit of development friction.

_includes/selector.html
jameslamb: I'm proposing just unconditionally dropping 12.2 for simplicity and to nudge people towards newer images.

But if reviewers disagree, I'd be happy to introduce a similar structure to what I'm proposing for Python versions, that allows the CUDA versions to differ by release. I think that'd be generically useful, just didn't add it here to limit the size of this PR.

jakirkham:
Thanks James! 🙏

This looks good

One thing that stood out is if Python 3.9 is selected for stable...

Screenshot 2024-08-27 at 11 54 12 AM

... and then nightly is picked, it still shows Python 3.9

Screenshot 2024-08-27 at 11 54 23 AM

Not sure if there is a way to fix this to move to a newer Python when selecting nightlies

Admittedly this might not be worth the trouble. Just wonder if it would cause confusion

jarmak-nv:
> ... and then nightly is picked, it still shows Python 3.9

Usually I'd workaround this by adding a bit of logic in the `releaseClickHandler` saying if they click nightly then update Python to 3.10 since it's in the same relative position (ie the lowest allowable version)

jameslamb:
I can add something like that.

jameslamb:
Ok, think I've handled that case: https://github.com/rapidsai/docs/pull/534/commits/e6073f3169996ba1a67ee627dc7d6cf01a6aa3b3

Try it out at https://deploy-preview-534--docs-rapids-ai.netlify.app/install

1. select `Stable (24.08)` on the Release line
2. select `3.9` on the Python line
3. select `Nightly (24.10a)` on the Release line

You should see the following:

* Python 3.9 grayed out, not selectable
* Python 3.10 selected
* `python=3.10` in the conda installation instructions

Tested the same thing with Docker, that looks like it's working as well.

I've written very little JavaScript, so I have no idea if what I'm doing is performant (and I know performance can matter for search rankings). Please do let me know if there are different data structures or patterns you think I should be using.

jakirkham (approved):
Thanks James! 🙏

LGTM

Let's see if Ben has any more thoughts 🙂

jarmak-nv:
The code changes lgtm, I like that you made a more permanent implementation vs one that would need to be removed like I suggested.

My only q - the deploy preview isn't working for me and it doesn't look like netlify is reporting an error: `Error code: PR_CONNECT_RESET_ERROR`

https://www.netlifystatus.com/

Has anyone been able to check it? Just want to be sure there's not something causing an issue in the selector.

jarmak-nv:
Looks like it's working for me now - (I'm on mac firefox, but yeah no issue now.)

LGTM - thanks @jameslamb

jarmak-nv (approved):

jakirkham:
Thanks James and Ben! 🙏

PR #536, opened 2024-09-03
Adding `nx-cugraph` to RAPIDS Install Selector Tool

nv-rliu:
Closes https://github.com/rapidsai/graph_dl/issues/604 & closes https://github.com/rapidsai/graph_dl/issues/609

In preparation for its stable release, this PR adds `NetworkX + nx-cugraph` as an option to the RAPIDS Install Selector Tool > Additional Packages. 

### Proposed changes:

- There _note_ also mentions that `nx-cugraph` is part of the Standard RAPIDS metapackage.
- Users now have the option to install `NetworkX + nx-cugraph` via the "Additional Packages" list when installing via `conda`

image

- `nx-cugraph` will be included in the Standard pip installation

image

## Notes for Reviewers

- The additional packages list is now sorted alphabetically (this can be reverted as necessary).
- When a user clicks "NetworkX + nx-cugraph", this only impacts the `conda` install option. This adds `networkx nx-cugraph` to the install command.


bdice (changes requested):
We have historically been very cautious about adding buttons to this list. We have far more packages in RAPIDS than we have room to list (we don't show RMM, KvikIO, ...). I would propose editing the cuGraph button to say `cuGraph / nx-cugraph` if that is a sufficient solution. You can see how we handle that logic for `cuSpatial/cuProj` here:

https://github.com/rapidsai/docs/blob/c09f11adf97b8a372f3cbac92e165288a6a19998/_includes/selector.html#L490-L494

rlratzel:
> I would propose editing the cuGraph button to say `cuGraph / nx-cugraph` if that is a sufficient solution. 

We discussed that, but decided that might not be desired since installing cugraph with nx-cugraph brings along a lot of extra dependencies that nx-cugraph does not use and users may not want (dask, cudf, etc.).

nx-cugraph is not in the RAPIDS meta-package at the moment either, so we're planning on adding it there too.  That will help and seems correct for a RAPIDS install, but it's not as noticeable as a dedicated button and doesn't help users that just want Accelerated NetworkX.

@bdice if nx-cugraph isn't button-worthy, is there an alternative we can discuss?

rlratzel:
> @bdice if nx-cugraph isn't button-worthy, is there an alternative we can discuss?

cc @beckernick for any additional suggestions.

bdice:
We could revisit the entire UX of the package selector. Maybe there’s a multi-row interface or something that uses a list with checkboxes that would still be visually appealing and provide more flexibility.

bdice:
Also, I don’t want to be a deciding voice here against adding a button. I just raise that concern because we have hesitated on similar changes in the past.

rlratzel:
Discussed offline with @bdice @beckernick @nv-rliu @aravenel @quasiben and decided that the best option would be to add a button to "Additional Packages" named "NetworkX + nx-cugraph" (or just "NetworkX" with a pop-up note indicating nx-cugraph is installed too?) which installs NetworkX and nx-cugraph.

_includes/selector.html
```diff
@@ -657,6 +658,9 @@
                     if (this.active_additional_packages.includes("TensorFlow")) {
                         notes = [...notes, "TensorFlow requires CUDA 12."];
                     }
+                    if (this.active_additional_packages.includes("NetworkX + nx-cugraph")) {
+                        notes = [...notes, "Selecting this option will install both standard NetworkX and a GPU-accelerated backend by RAPIDS"];
+                    }
```
nv-rliu, line 663: Any feedback on this messaging?
Preview:
image

_includes/selector.html
```diff
@@ -392,9 +392,9 @@
             packages: ["Standard", "Choose Specific Packages"],
             additional_pip_packages: ["cuDF", "dask-cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuCIM", "RAFT", "cuVS"],
             additional_rapids_packages: ["cuDF", "cuML", "cuGraph", "cuSpatial/cuProj", "cuxfilter", "cuCIM", "RAFT", "cuVS"],
-            additional_packages: ["JupyterLab", "Plotly Dash", "Graphistry", "TensorFlow", "Xarray-Spatial", "PyTorch"],
+            additional_packages: ["Graphistry", "JupyterLab", "NetworkX + nx-cugraph", "Plotly Dash", "PyTorch", "TensorFlow", "Xarray-Spatial"],
             note_prefix: "<i class='fas fa-info-circle text-blue'></i>",
-            rapids_meta_pkgs: ["cuDF", "cuML", "cuGraph", "cuSpatial", "cuProj", "cuxfilter", "cuCIM", "RAFT"],
+            rapids_meta_pkgs: ["cuDF", "cuML", "cuGraph", "cuSpatial", "cuProj", "cuxfilter", "cuCIM", "nx-cugraph", "RAFT"],
```
bdice, line 397: We don't need both `NetworkX + nx-cugraph` and a separate button for `nx-cugraph`. Choose one or the other.

_includes/selector.html
```diff
@@ -566,13 +567,15 @@
...
+                    var additional_pip_commands = libraryToPkg("nx-cugraph");
+                    pkgs = pkgs.concat(additional_pip_commands);
                 } else {
                     // sort active_packages to appear in the same order as the additional_pip_packages list
                     this.active_packages.sort((a, b) => this.additional_pip_packages.indexOf(a) - this.additional_pip_packages.indexOf(b));
                     var pkgs = this.active_packages
                         .filter(pkg => pkg !== "Choose Specific Packages")
                         .flatMap(libraryToPkg);
-                }
+                    }
```
bdice, line 578: Revert this.
```suggestion
                }
```

_includes/selector.html
```diff
@@ -657,6 +660,9 @@
                     if (this.active_additional_packages.includes("TensorFlow")) {
                         notes = [...notes, "TensorFlow requires CUDA 12."];
                     }
+                    if (this.active_additional_packages.includes("NetworkX + nx-cugraph")) {
+                        notes = [...notes, "Selecting this option will install both standard NetworkX and a GPU-accelerated backend by RAPIDS"];
```
bdice, line 664: I don't think we need notes here, the button label "NetworkX + nx-cugraph" is clear enough to me.

_includes/selector.html
bdice: Sorry, maybe I misunderstood this. I need to look at where this variable is used.

_includes/selector.html
```diff
@@ -457,6 +457,7 @@
             },
             getAdditionalPkgName(pkg) {
                 var pkg_names = {
+                    "NetworkX + nx-cugraph": "networkx nx-cugraph",
```
bdice, line 460: We need a version pinning on nx-cugraph. All the RAPIDS packages get `=24.08` or `=24.10`. Copy that here if nx-cugraph is versioned the same way.
```suggestion
                    "NetworkX + nx-cugraph": "networkx nx-cugraph",
```

_includes/selector.html
bdice: I get it now. The `rapids_meta_pkgs` is a list of things in the `rapids` conda metapackage. Apologies, you did this right. I just didn't understand it at first.

_includes/selector.html
rlratzel: I confirmed that the RAPIDS metapackage includes nx-cugraph, so do we still need to add `nx-cugraph` to the install line if it also includes `rapids`?

_includes/selector.html
rlratzel: I agree with @bdice here after seeing it.  

Although, another option which is closer to a suggestion Nick made earlier and addresses a comment I made above, would be to change the button to just say "NetworkX" and then make the note say:

> This option installs NetworkX which will be GPU accelerated by the nx-cugraph backend included in RAPIDS.

Although I don't feel strongly either way, and perhaps just having the button say "NetworkX + nx-cugraph" is adequate and not confusing.

_includes/selector.html
nv-rliu: This doesn't add a separate button for `nx-cugraph`. This is for the disclaimer that tells users that the Standard RAPIDS metapackages comes with `nx-cugraph` 

image

_includes/selector.html
bdice: Technically we do not. We only need it if we're selecting individual packages, as the `rapids` metapackage would transitively include `nx-cugraph`. However, it'd be an unexpected UX for a button labeled "NetworkX + nx-cugraph" to not add `nx-cugraph` explicitly. I would leave the behavior as-is.

nv-rliu: Version pinning changes for both Stable and Nightly:

image

image

rlratzel (approved):
Nice!

_includes/selector.html
bdice, line 397: Let's put `nx-cugraph` next to `cuGraph`. It is more related to cuGraph than cuCIM or RAFT. Also we happen to be missing cuVS from this list, so I fixed that (I used the same order as the buttons except for nx-cugraph because it doesn't have a package button, but I think this should be fine).
```suggestion
            rapids_meta_pkgs: ["cuDF", "cuML", "cuGraph", "nx-cugraph", "cuSpatial", "cuProj", "cuxfilter", "cuCIM", "RAFT", "cuVS"],
```

_includes/selector.html
```diff
@@ -456,7 +456,10 @@
                 };
             },
             getAdditionalPkgName(pkg) {
+                // used to pin the version of nx-cugraph for conda install
+                var rapids_version = this.active_release === "Stable" ? this.getStableVersion() : this.getNightlyVersion();
                 var pkg_names = {
+                    "NetworkX + nx-cugraph": `networkx nx-cugraph=${rapids_version}`,
```
bdice, line 462: This is annoying... the nx-cugraph version pinning is highlighted but the normal version pins are not. 
image

I'm willing to accept this as-is but we could refactor the block below so it only highlights the package name and not the pinning:
```javascript
if (this.active_additional_packages.length) {
    add_pkgs_ln = indentation + this.active_additional_packages
    .map(pkg => this.highlightCondaPkg(this.getAdditionalPkgName(pkg)))
        .join(" ");
}
```

bdice (approved):
Let’s merge this. But it is missing a key feature: people who select “pip” and choose specific packages do not have an option to pick nx-cugraph.

_includes/selector.html
nv-rliu: @bdice Would it be alright to file an issue for this and refactor the function that highlights package names in a PR down the road?

bdice: Yes -- that's totally fine.

nv-rliu:
> Let’s merge this. But it is missing a key feature: people who select “pip” and choose specific packages do not have an option to pick nx-cugraph.

@bdice Do you think that this key feature should be added my also including an "Additional Packages" bar for `pip` installation?

PS. Is there a reason why this isn't included for `pip` when it was added to `conda`?

bdice:
We've just never implemented a pip "additional packages". I think we could open discussion for that with the team internally and decide how to proceed. In the short term we probably only need a solution for adding nx-cugraph.

PR #538, opened 2024-09-05
Add `nx-cugraph` Intro to API Docs Page

nv-rliu:
Closes https://github.com/rapidsai/graph_dl/issues/606

### Proposed Changes:
 - This PR adds a section to RAPIDS Docs > API Docs for `nx-cugraph`
 - The `nx-cugraph` docs will live inside the cugraph docs, but they have their own dedicated landing page and Table of Contents (https://github.com/rapidsai/cugraph/pull/4669)

### Proposed change looks like:

image
as of [`5458bbd`](https://github.com/rapidsai/docs/pull/538/commits/5458bbd731e42211aef44364e0f8b278ad723457)


_data/docs.yml
```diff
@@ -46,6 +46,17 @@ apis:
       legacy: 1
       stable: 1
       nightly: 1
+  nx-cugraph:
+    name: nx-cugraph
+    path: cugraph # TODO: determine where to point nx-cugraph api docs
+    desc: 'nx-cugraph is a NetworkX backend that provides GPU acceleration to many popular NetworkX algorithms. With nx-cugraph, users can have GPU-based, large-scale performance without needing to change their pre-existing NetworkX code.'
+    ghlink: https://github.com/rapidsai/nx-cugraph
```
bdice, line 53: This is an internal repository link and won't work for public users. Just use the cuGraph link since nx-cugraph lives in the same repo. (This is how we handle dask-cudf above).
```suggestion
    ghlink: https://github.com/rapidsai/cugraph
```

_data/docs.yml
```diff
@@ -46,6 +46,17 @@ apis:
...
+    cllink: https://github.com/rapidsai/nx-cugraph/blob/main/CHANGELOG.md
```
bdice, line 54: As above:
```suggestion
    cllink: https://github.com/rapidsai/cugraph/blob/main/CHANGELOG.md
```

_data/docs.yml
```diff
@@ -46,6 +46,17 @@ apis:
...
+    versions:
+      # enable or disable links; 0 = disabled, 1 = enabled
+      legacy: 0
+      stable: 0
```
bdice, line 58: At this point we do have legacy and stable docs for nx-cugraph because it has existed since 24.06.

Legacy: https://docs.rapids.ai/api/cugraph/legacy/nx_cugraph/
Stable: https://docs.rapids.ai/api/cugraph/stable/nx_cugraph/
```suggestion
      legacy: 1
      stable: 1
```

_data/docs.yml
```diff
@@ -46,6 +46,17 @@ apis:
       legacy: 1
       stable: 1
       nightly: 1
+  nx-cugraph:
+    name: nx-cugraph
+    path: cugraph # TODO: determine where to point nx-cugraph api docs
```
bdice, line 51: The docs are hosted as part of the cugraph docs, so this is correct.
```suggestion
    path: cugraph
```

_data/docs.yml
nv-rliu: I was under the impression that we would be migrating `nx-cugraph` out of cugraph as a part of the 24.10 release.

rlratzel: We can change this to use the cugraph repo for now and then just submit a followup PR when the nx-cugraph sources move to the new nx-cugraph repo later.

bdice (commented):
Does it make sense to merge this before nx-cugraph has its own documentation? Should we just note nx-cugraph under the cuGraph section? All the other entries on this page have a unique `path:` which corresponds to the docs URL. We don't publish nx-cugraph docs on their own (yet) so I don't see a large benefit to adding this link because it duplicates the link to existing cuGraph content.

I would vote to wait on merging this until we publish nx-cugraph docs at their own page (https://docs.rapids.ai/api/nx-cugraph/stable/, which doesn't exist yet) and then we can update the `path:` accordingly.

nv-rliu:
This should point to the pages created by this PR https://github.com/rapidsai/cugraph/pull/4669

PR #546, opened 2024-10-03
Add redirect for /deployment to /deployment/stable

bdice:
https://docs.rapids.ai/deployment should point to https://docs.rapids.ai/deployment/stable/


jameslamb (approved):
TIL that that we have this redirect mechanism, thank you! Totally support this.

ajschmidt8 (approved):

jacobtomlinson:
Hmm I'm surprised we needed this, we used to have this in place. I wonder when it got removed.

PR #559, opened 2024-10-29
Declare Linux support via glibc versions

bdice:
This PR clarifies that RAPIDS is supported on Linux distributions with glibc >= 2.28. This is more inclusive than limiting to only the distributions in our CI matrix, as there are many flavors of Linux that are functionally similar for the needs of RAPIDS.

### DIFF
--- a/install/index.md
+++ b/install/index.md
@@ -68,13 +68,13 @@ If conda has incorrectly identified the CUDA driver, you can [override by settin
 ### **Docker Issues**
 <i class="fas fa-exclamation-triangle"></i> RAPIDS `23.08` brought significant Docker changes. <br/>
 To learn more about these changes, please see the [RAPIDS Container README](https://hub.docker.com/r/rapidsai/base){: target="_blank"}. Some key notes below:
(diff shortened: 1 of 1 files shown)


install/index.md
```diff
@@ -283,7 +289,7 @@ Windows users can now tap into GPU accelerated data science on their local machi
 
 ### **WSL2 Additional Prerequisites**
 
-<i class="fas fa-desktop text-white"></i> **OS:** Windows 11 with Ubuntu 22.04 instance for WSL2. <br/>
+<i class="fas fa-desktop text-white"></i> **OS:** Windows 11 with a WSL2 installation of Ubuntu (minimum version 20.04). <br/>
```
jakirkham, line 292: Looks like this was Ubuntu 22.04 and then became 20.04. Was this meant to still be 22.04?

```suggestion
<i class="fas fa-desktop text-white"></i> **OS:** Windows 11 with a WSL2 installation of Ubuntu (minimum version 22.04). <br/>
```

bdice: This should be Ubuntu 20.04. It was incorrect before.

PR #561, opened 2024-11-13
Quote packages so shells don't get unhappy

### LINKED ISSUE #560: [BUG] Install command with `*` in it leads to shell error on zsh
**Describe the bug**
When I copy and paste an install command from the "Install selector" I get the following error in zsh:
```
pip install \
    --extra-index-url=https://pypi.nvidia.com \
    cudf-cu12==24.10.*
zsh: no matches found: cudf-cu12==24.10.*
```
this is from a selection like this:
Image

I reckon that quoting the `cudf-cu12==24.10.*` bit works for all shells and does no harm even when it isn't needed

**To Reproduce**
1. [Go to '...'](https://docs.rapids.ai/install/)
2. Configure as in the screenshot above (pip, cudf)
3. Copy and paste into zsh

**Expected behavior**
It "just works"

**Screenshots**

betatim:
This attempts to fix #560

This moves the quoting to the end of `getpipCmdHtml` and always applies it. Previously only the nightly version got quoted. but it required a leading `"` in each package name and a closing `"` in the CUDA suffix.

I clicked through a few combinations of things for the `pip` option. As far as I can tell it works for all of them

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -586,6 +586,8 @@
                     pkgs.push(["\"dask-cuda" + cuda_suffix.slice(5)])
                 }
 
+                pkgs = pkgs.flatMap(pkg => '"' + pkg + '"');
+
                 // For every n packages add a new line with a "\" character
                 // We need i += n + 1 since the splice adds a new element to the array
                 if (this.active_release === "Stable") {
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -556,14 +556,14 @@
                 else {
                     index_url = `--${this.highlightFlag("extra-index-url")}=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple`;
                     var version = this.removeLeadingZeros("{{ site.data.releases.nightly.version }}")
-                    cuda_suffix = cuda_suffix + `>=${version}.0a0,<=${version}"`;
+                    cuda_suffix = cuda_suffix + `>=${version}.0a0,<=${version}`;
                     var libraryToPkg = (pkg) => {
                         pkg = pkg.toLowerCase();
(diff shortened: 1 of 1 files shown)


_includes/selector.html
```diff
@@ -556,14 +556,14 @@
...
-                        if (pkg === "cuspatial/cuproj") return ["\"cuspatial" + cuda_suffix, "\"cuproj" + cuda_suffix];
-                        if (pkg === "raft") return ["\"pylibraft" + cuda_suffix, "\"raft-dask" + cuda_suffix];
-                        if (pkg === "cuvs") return ["\"cuvs" + cuda_suffix, "\"pylibraft" + cuda_suffix];
-                        if (pkg === "cugraph/nx-cugraph") return ["\"cugraph" + cuda_suffix, "\"nx-cugraph" + cuda_suffix];
-                        return ["\"" + pkg + cuda_suffix];
+                        if (pkg === "cuspatial/cuproj") return ["cuspatial" + cuda_suffix, "cuproj" + cuda_suffix];
+                        if (pkg === "raft") return ["pylibraft" + cuda_suffix, "raft-dask" + cuda_suffix];
+                        if (pkg === "cuvs") return ["cuvs" + cuda_suffix, "pylibraft" + cuda_suffix];
+                        if (pkg === "cugraph/nx-cugraph") return ["cugraph" + cuda_suffix, "nx-cugraph" + cuda_suffix];
+                        return ["" + pkg + cuda_suffix];
```
bdice, line 566: Is this empty string needed?
```suggestion
                        return [pkg + cuda_suffix];
```

bdice (commented):
One small question.

bdice (approved):

PR #564, opened 2024-12-13
add `ci` to options for label checker categories

gforsyth:
I opened a PR over on `dask-cuda` and didn't see a `ci` label listed as an option to pass the `label-checker`, but it seems like it does exist.

Not sure if it's universal across RAPIDS repos, but if it is, it should be added to this page.

### DIFF
--- a/resources/label-checker.md
+++ b/resources/label-checker.md
@@ -34,6 +34,7 @@ In order for the _Label Checker_ check to pass, pull requests must include **one
 
 - `category` - categorizes the pull request
   - `bug`
+  - `ci`
   - `doc`
   - `feature request`
   - `improvement`


bdice (changes requested):
The allowed list of labels is here: https://github.com/rapidsai/ops-bot/blob/fdeb43f78722173f39944d331ac36cff52594c61/src/plugins/LabelChecker/label_checker.ts#L61

`ci` is an informational label but not a permitted category, so the label checker will fail. This category information is used to generate changelogs.

Every RAPIDS library PR must be one marked with one of the listed categories, and breaking or non-breaking.

gforsyth:
Ahh, ok, thanks for clarifying.  Do you think adding a `ci` category makes sense?

bdice:
> Ahh, ok, thanks for clarifying. Do you think adding a `ci` category makes sense?

Usually we just put build-related changes in “improvement.” Calling out CI changes in the changelog is probably not that useful, as it won’t affect library users directly.

PR #569, opened 2025-01-23
Enable librmm download and publishing

AyodeAwe:
`librmm` docs were previously [removed](https://github.com/rapidsai/docs/pull/517) but now need to be added back because crosslinks from `libcudf` currently `404`. See [here](https://github.com/rapidsai/docs/pull/517#issuecomment-2221045326).

The `librmm` entry will not be listed in https://docs.rapids.ai/api/ but will continue to be deployed to the website so that crosslinks from `libcudf` can continue to function.

### DIFF
--- a/_includes/api-docs.html
+++ b/_includes/api-docs.html
@@ -1,5 +1,6 @@
 {% for lib in include.data %}
 {% assign api = lib[1] %}
+{% if api.name != "librmm" %}
 {% comment %} HACK: below operation returns an array of stable/nightly/legacy values only if they're enabled in docs.yml. see issue #97 {% endcomment %}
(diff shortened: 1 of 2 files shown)


_includes/api-docs.html
```diff
@@ -1,10 +1,12 @@
 {% for lib in include.data %}
 {% assign api = lib[1] %}
+{% if api.name != "librmm" %}
```
bdice, line 3: Can we instead add a field like `if not api.hidden` and set `hidden: true` in the YAML?

AyodeAwe: Good catch. Fixed. Thanks for that review!

ajschmidt8 (approved):

PR #570, opened 2025-01-27
RSN for conda package `legacy` labelling

gforsyth:
RSN for moving select conda packages behind a new `legacy` label concurrent with the `v25.04` release.

I wasn't sure whether to note the `notice_rapids_version` as the version where
this change takes place (`25.04`) or the versions impacted by the change
(`<=23.02`). I think `<=23.02` makes more sense and I've written it as such, but
happy to change things around.

xref rapidsai/build-planning#84


_notices/rsn0042.md
```diff
@@ -0,0 +1,60 @@
...
+notice_created: 2025-01-27
+# 'notice_updated' should match 'notice_created' until an update is made
+notice_updated: 2025-01-27
+---
+
+## Overview
+
+RAPIDS will move certain older conda packages in the `rapidsai` conda channel
+under a new label: `legacy`. These packages will still be available by appending
+`--label legacy` to installation and search commands.
```
bdice, line 32: I'm a little unclear on this. It looks like `--label` is used when uploading, but for users to install, they need something like `--channel rapidsai --channel rapidsai/label/legacy`. Does that sound correct?

See https://docs.anaconda.com/anacondaorg/user-guide/work-with-labels/#testing-the-discoverability-of-the-package

_notices/rsn0042.md
```diff
@@ -0,0 +1,60 @@
...
+e.g. if you are creating an environment like:
+
+```python
+mamba create -n rapids-23.02 rapids=23.02 python=3.10 'cuda-version>=11.4,<=11.8' -c rapidsai -c conda-forge -c nvidia
+```
+
+you will need to instead run
+
+```python
+mamba create -n rapids-23.02 rapids=23.02 python=3.10 'cuda-version>=11.4,<=11.8' -c rapidsai -c conda-forge -c nvidia --label legacy
```
bdice, line 59: I expected this to read:
```suggestion
mamba create -n rapids-23.02 rapids=23.02 python=3.10 'cuda-version>=11.4,<=11.8' -c rapidsai -c rapidsai/label/legacy -c conda-forge -c nvidia

```

_notices/rsn0042.md
gforsyth: Yup, thanks for catching that @bdice !

_notices/rsn0042.md
```diff
@@ -0,0 +1,60 @@
...
+---
+
+## Overview
+
+RAPIDS will move certain older conda packages in the `rapidsai` conda channel
+under a new label: `legacy`. These packages will still be available by adding
+`-c rapidsai/label/legacy` to installation and search commands.
+
+By moving these packages out of the main portion of the channel, newer RAPIDS
+conda environments will be solvable with `strict_channel_priority`, speeding up
```
bdice, line 35:
```suggestion
conda environments will be solvable with `channel_priority: strict`, speeding up

```

_notices/rsn0042.md
```diff
@@ -0,0 +1,60 @@
...
+environment solve times.
+
+All RAPIDS conda install commands for RAPIDS versions `v23.04+` will continue to
+work without any user intervention.
+
+This change will occur concurrently with the release of RAPIDS `v25.04`.
```
bdice, line 41: Just checking: We will be able to test strict channel priority in our CI before the 25.04 release, right? I want to be sure we minimize the risk of breaking recent releases (25.02 / 25.04) during this change.

bdice (approved):
Looks good! One small suggestion and one question.

_notices/rsn0042.md
gforsyth: We'll be able to test `25.02` once all of those packages are released using the same method I tested the previous releases with.  Regarding 25.04, we can also release all of those packages, test, then move the selected older packages under the `legacy` label, or we can move them under the `label`, see if it works, and then undo the labeling if it breaks.  

If `25.02` works, I'd feel confident moving the packages straight to the `label` for 25.04, and that would let us accurately advertise strict channel priority for the new install instructions.

_notices/rsn0042.md
```diff
@@ -32,7 +32,8 @@ under a new label: `legacy`. These packages will still be available by adding
 `-c rapidsai/label/legacy` to installation and search commands.
 
 By moving these packages out of the main portion of the channel, newer RAPIDS
-conda environments will be solvable with `strict_channel_priority`, speeding up
+conda environments will be solvable with `channel_priority: strict`, speeding up
+
```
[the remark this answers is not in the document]
gforsyth: thanks, @bdice !  yeah, I've run into this before, too, and I've not figured it out

_notices/rsn0042.md
bdice: That's helpful. What about enabling strict priority in our CI builds/tests, though? It sounds like you're discussing the user experience that we expect after the release is complete.

gforsyth: We can't really enable them for CI until we move the older packages under `legacy`.  The `conda-metachannel` package is quite brittle at this point and I don't think we want that spun up as part of our CI

bdice: Does that mean we will make the `legacy` label along with the 25.04 release, and then enable strict priority in our CI during development for 25.06?

gforsyth: Yes

jameslamb (changes requested):
Coming here to review based on @jakirkham 's suggestion in https://github.com/rapidsai/build-planning/issues/14#issuecomment-2696014916

I'm leaving a blocking review because a couple of these have to be fixed (like the RSN number).

Can you also please update this to latest `main`? There are some new commits there since the last time this was modified.

_notices/rsn0042.md
bdice: You're asking a good question. I think the only confusing part here is timing.

We can't enable strict channel priority until we move "legacy" packages (needed for releases prior to 23.04) to the `legacy` label. There are packages like `datashader` that are implicated here.

We plan to move things to the `legacy` label _when 25.04 is released_. Then we can enable strict priority in 25.06.

However, there is a path where we enable strict priority _gradually in each repo's CI_, which I think we could start in 25.04.

gforsyth: > Why is this? I'd think that in RAPIDS CI for 25.04, the pins like ==25.4.* we have for all RAPIDS packages would be sufficient to allow this.

channel priority doesn't consider versions -- if there is ANY version of a package in a higher priority channel, that is the only package that gets considered.  because we have (especially) a patched version of `datashader` (although there are other problem packages) in `rapidsai`, most of our solves won't work because it'll see `datashader` in `rapidsai` but the single version in `rapidsai` won't satisfy the current dependency requirements and the solve will fail.

jameslamb: ahhhhhhh ok, thank you both for explaining. Ok then this makes sense to me as-is.

Agree with @bdice that it'd be really helpful to try to enable this repo-by-repo within the 25.04 cycle, to hopefully prevent accidental introduction of even *more* things that complicate the move to strict channel priority. And because it's a solution to the "did not install the package built in CI" problem: https://github.com/rapidsai/build-planning/issues/14#issuecomment-2659761340

jameslamb (approved):
Changes look good to me and all my concerns have been addressed. Thank you for moving this forward!

_notices/rsn0044.md
```diff
@@ -0,0 +1,60 @@
...
+#   "Proposal" - "blue"
+#   "Completed" - "green"
+#   "Review" - "purple"
+#   "In Progress" - "yellow"
+#   "Closed" - "red"
+notice_topic: Platform Support Change
+notice_rapids_version: "<=v23.02"
+notice_created: 2025-03-04
+# 'notice_updated' should match 'notice_created' until an update is made
+notice_updated: 2025-03-04
```
bdice, line 25:
```suggestion
notice_created: 2025-04-09
# 'notice_updated' should match 'notice_created' until an update is made
notice_updated: 2025-04-09

```

bdice (approved):
Looks good to me. This has two approvals and needs to be published with the 25.04 release so I will merge it.

jakirkham (approved):
Thanks Gil for all of your hard work here and everyone who helped review! 🙏

Looking forward to having strict channel priority available for users in RAPIDS 25.04 🥳

_notices/rsn0044.md
```diff
@@ -0,0 +1,60 @@
...
+nav_exclude: true
+notice_type: rsn
+# Update meta-data for notice
+notice_id: 44 # should match notice number
+notice_pin: true # set to true to pin to notice page
+
+title: "Moving older RAPIDS conda packages to a `legacy` label"
+notice_author: RAPIDS Ops
+notice_status: Proposal
+notice_status_color: blue
```
bdice, line 14: Needs to be yellow since we plan to do this but haven't completed it yet. We don't really use "Proposal."
```suggestion
notice_status: In Progress
notice_status_color: yellow
```

bdice (approved):

PR #577, opened 2025-02-14
update docs on using CI artifacts across PRs

jameslamb:
During the 25.02 release, I worked through a bunch of PRs that relied on other PRs' wheels and conda packages, sometimes in a dependency chain 3 or 4 libraries deep.

This proposes some changes to the docs on how to do that. In general... instead of manually modifying every `ci/` script, it's possible to put the changes into a single `ci/use_{conda_packages,wheels}_from_prs.sh` that is then `source`'d into all scripts, reducing duplication and mistakes.

### DIFF
--- a/resources/github-actions.md
+++ b/resources/github-actions.md
@@ -252,7 +252,7 @@ LIBRAFT_WHEELHOUSE=$(
   RAPIDS_PY_WHEEL_NAME="libraft_${RAPIDS_PY_CUDA_SUFFIX}" rapids-get-pr-wheel-artifact raft 2433 cpp
 )
 
-# write a pip constraints file saying e.g. "whenever you encounter a requirement for 'librmm-cu12', use this file"
+# write a pip constraints file saying e.g. "whenever you encounter a requirement for 'librmm-cu12', use this wheel"
 cat > /tmp/constraints.txt <<EOF
 librmm-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${LIBRMM_WHEELHOUSE}/librmm_*.whl)
 libraft-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${LIBRAFT_WHEELHOUSE}/libraft_*.whl)
@@ -292,7 +292,7 @@ RMM_WHEELHOUSE=$(
   RAPIDS_PY_WHEEL_NAME="rmm_${RAPIDS_PY_CUDA_SUFFIX}" rapids-get-pr-wheel-artifact rmm 1678 python
 )
 
-# write a pip constraints file saying e.g. "whenever you encounter a requirement for 'librmm-cu12', use this file"
+# write a pip constraints file saying e.g. "whenever you encounter a requirement for 'librmm-cu12', use this wheel"
 cat > /tmp/constraints.txt <<EOF
 libkvikio-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${LIBKVIKIO_WHEELHOUSE}/libkvikio_*.whl)
 librmm-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${LIBRMM_WHEELHOUSE}/librmm_*.whl)
--- a/resources/github-actions.md
+++ b/resources/github-actions.md
@@ -176,51 +176,144 @@ The example code below demonstrates building and testing with conda packages fro
 Replace the pull request numbers and library names as needed.
 Remember that changes to use CI artifacts should be _temporary_ and should be reverted prior to merging any required changes in that PR.
 
-**Example 1:** Building `libcuml` (C++) using `librmm`, `libraft`, `libcumlprims_mg` PR artifacts.
+**Example 1:** Building `libcuml` (C++) using `librmm` and `libraft` PR artifacts.
 
-```sh
-# ci/build_cpp.sh
+Add a new file called `ci/use_conda_packages_from_prs.sh`.
 
-LIBRMM_CHANNEL=$(rapids-get-pr-conda-artifact rmm 1095 cpp)
+```shell
+# ci/use_conda_packages_from_prs.sh
+
+# download CI artifacts
 LIBRAFT_CHANNEL=$(rapids-get-pr-conda-artifact raft 1388 cpp)
-LIBCUMLPRIMS_CHANNEL=$(rapids-get-pr-conda-artifact cumlprims_mg 129 cpp)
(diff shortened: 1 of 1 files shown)


jameslamb:
I don't have permissions to add reviewers in this repo, so tagging a few folks I expect would be interested:

@vyasr @bdice @gforsyth @KyleFromNVIDIA @msarahan @divyegala

msarahan (approved):
Big improvement! Thanks for your work on this.

divyegala (commented):
This is a very needed improvement. I was not aware of how easy it can be to use artifacts from other PRs, and it seems to me the right way to do it than updating `get_raft.cmake` in cuml, for example.

vyasr:
Thanks James! I forgot to follow up on this request with you, so thanks for remembering and following up. I have no review/merge powers here so I'll let @bdice take a look since he does.

resources/github-actions.md
```diff
@@ -176,51 +176,144 @@ The example code below demonstrates building and testing with conda packages fro
...
+# ci/use_conda_packages_from_prs.sh
+
+# download CI artifacts
 LIBRAFT_CHANNEL=$(rapids-get-pr-conda-artifact raft 1388 cpp)
-LIBCUMLPRIMS_CHANNEL=$(rapids-get-pr-conda-artifact cumlprims_mg 129 cpp)
+LIBRMM_CHANNEL=$(rapids-get-pr-conda-artifact rmm 1095 cpp)
+
+# make sure they can be found locally
+conda config --system --add channels "${LIBRAFT_CHANNEL}"
+conda config --system --add channels "${LIBRMM_CHANNEL}"
```
bdice, line 192: Note: this will change with rattler-build. Should we preface this with that disclaimer?

resources/github-actions.md
```diff
@@ -176,51 +176,144 @@ The example code below demonstrates building and testing with conda packages fro
...
+# write a pip constraints file saying e.g. "whenever you encounter a requirement for 'librmm-cu12', use this wheel"
+cat > /tmp/constraints.txt <<EOF
+librmm-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${LIBRMM_WHEELHOUSE}/librmm_*.whl)
+libraft-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${LIBRAFT_WHEELHOUSE}/libraft_*.whl)
+EOF
+
+export PIP_CONSTRAINT=/tmp/constraints.txt
+```
+
+Then copy the following into every script in the `ci/` directory that is doing `pip` installs or wheels builds with e.g. `pip wheel`.
```
bdice, line 264:
```suggestion
Then copy the following into every script in the `ci/` directory that is doing `pip` installs or wheel builds with e.g. `pip wheel`.
```

resources/github-actions.md
```diff
@@ -176,51 +176,144 @@ The example code below demonstrates building and testing with conda packages fro
...
+
+```shell
+source ./ci/use_wheels_from_prs.sh
+```
+
+This should generally be enough.
+If any of the other CI scripts are already setting environment variable `PIP_CONSTRAINT`, you may need to
```
bdice, line 271:
```suggestion
If any of the other CI scripts are already setting the environment variable `PIP_CONSTRAINT`, you may need to
```

resources/github-actions.md
```diff
@@ -176,51 +176,144 @@ The example code below demonstrates building and testing with conda packages fro
...
+)
+RMM_WHEELHOUSE=$(
+  RAPIDS_PY_WHEEL_NAME="rmm_${RAPIDS_PY_CUDA_SUFFIX}" rapids-get-pr-wheel-artifact rmm 1678 python
+)
+
+# write a pip constraints file saying e.g. "whenever you encounter a requirement for 'librmm-cu12', use this wheel"
+cat > /tmp/constraints.txt <<EOF
+libkvikio-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${LIBKVIKIO_WHEELHOUSE}/libkvikio_*.whl)
+librmm-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${LIBRMM_WHEELHOUSE}/librmm_*.whl)
+rmm-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${RMM_WHEELHOUSE}/librmm_*.whl)
```
bdice, line 299:
```suggestion
rmm-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${RMM_WHEELHOUSE}/rmm_*.whl)
```

resources/github-actions.md
```diff
@@ -176,51 +176,144 @@ The example code below demonstrates building and testing with conda packages fro
...
+EOF
+
+export PIP_CONSTRAINT=/tmp/constraints.txt
+```
+
+Then copy the following into every script in the `ci/` directory that is doing `pip` installs or wheels builds with e.g. `pip wheel`.
```
bdice, line 305:
```suggestion
Then copy the following into every script in the `ci/` directory that is doing `pip` installs or wheel builds with e.g. `pip wheel`.
```

resources/github-actions.md
```diff
@@ -176,51 +176,144 @@ The example code below demonstrates building and testing with conda packages fro
...
+
+```shell
+source ./ci/use_wheels_from_prs.sh
+```
+
+As above, if any of the other CI scripts are already setting environment variable `PIP_CONSTRAINT`, you may need to
```
bdice, line 311:
```suggestion
As above, if any of the other CI scripts are already setting the environment variable `PIP_CONSTRAINT`, you may need to
```

bdice (approved):
I have a few small suggestions, but I'm approving this so that it can be merged once feedback is addressed.

resources/github-actions.md
jameslamb: I'd written this before I knew that: https://github.com/rapidsai/gha-tools/pull/146#issuecomment-2674948061

I don't think we should add a disclaimer like that. A note of the form "this won't work for builds using `rattler-build`" isn't something that folks reading this could do much with, by themselves.

It'd be better to add "update the docs on using CI artifacts" as part of the work for https://github.com/rapidsai/build-planning/issues/47

resources/github-actions.md
jameslamb: I don't think the form without "the" is grammatically incorrect, but I'll commit this because I can't think of any reason to choose one form over the other.

bdice (approved):

PR #580, opened 2025-02-26
Update docs to use rapids-conda-retry build and single-step environment solve.

bdice:
Addresses updates from https://github.com/rapidsai/build-planning/issues/149 and https://github.com/rapidsai/build-planning/issues/22.

### DIFF
--- a/resources/github-actions.md
+++ b/resources/github-actions.md
@@ -196,6 +196,9 @@ rapids-conda-retry build \
 
 **Example 2:** Testing `cudf` (Python) using `librmm`, `rmm`, and `libkvikio` PR artifacts.
 
+Here, `--file-key ${FILE_KEY}` must include dependencies on all the packages that were built in this PR and the packages being pulled from other PRs' artifacts.
+Thus, prepending these channels will ensure that the packages are pulled from the PR artifacts rather than `rapidsai-nightly`.
+
 ```sh
 # ci/test_python_common.sh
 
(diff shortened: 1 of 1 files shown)


bdice (commented):
@jakirkham Could you review this?

resources/github-actions.md
```diff
@@ -204,14 +204,18 @@ RMM_CHANNEL=$(rapids-get-pr-conda-artifact rmm 1223 python)
...
-  --channel "${CPP_CHANNEL}" \
-  --channel "${PYTHON_CHANNEL}" \
-  --channel "${LIBRMM_CHANNEL}" \
-  --channel "${LIBKVIKIO_CHANNEL}" \
-  --channel "${RMM_CHANNEL}" \
-  cudf libcudf
+# ...
+rapids-dependency-file-generator \
+  --output conda \
+  --file-key "${FILE_KEY}" \
```
jakirkham, line 210: Do we want to explain how this connects to the `dependencies.yaml` key?

Think the detail about this key including the packages produced in the previous build step would be interesting to readers

jakirkham (commented):
Thanks Bradley! 🙏

Had a question below

resources/github-actions.md
bdice: Done in f6156f3.

bdice:
#577 addressed these considerations in a different way. We can close this.

PR #591, opened 2025-04-10
Revert "Temporarily disable 25.06 cugraph docs"

raydouglass:
Reverts rapidsai/docs#590


bdice (approved):
This should be good now that cugraph-docs passed with this change: https://github.com/rapidsai/cugraph-docs/pull/107

raydouglass:
I've confirmed that `25.06` docs are now in the S3 bucket, so this should be safe to merge.

bdice (approved):

PR #596, opened 2025-04-25
Add RSN for cuGraph-DGL Removal

alexbarghi-nv:
Adds RSN for cuGraph-DGL removal after release 25.06.


bdice (approved):
This looks good to me. I don't see cuGraph-DGL in the `rapids` metapackages or Docker images, so I think the "Impact" section is complete.

aravenel (approved):

PR #597, opened 2025-04-30
Pin cuml-cpu drop notice

bdice:
We usually pin notices for dropping library/platform support but didn't pin this one for cuml-cpu. This PR adds it to the pinned notices.

Also I unpinned notices for releases prior to 2024.


jakirkham (approved):

csadorf (approved):
Approving RSN 43.

PR #600, opened 2025-05-06
Disable cuSpatial nightly packages and docs.

bdice:
cuSpatial will not be published in 25.06: https://docs.rapids.ai/notices/rsn0045/

This PR drops cuSpatial from the nightly package selector and nightly docs.

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -803,6 +803,11 @@
                if (!supported_docker_cuda_versions.includes(this.active_docker_cuda_ver)) {
                    this.active_docker_cuda_ver = supported_docker_cuda_versions[supported_docker_cuda_versions.length - 1];
                }
+
+               /* Remove cuSpatial/cuProj from active packages if in Nightly */
+               if (this.active_release === "Nightly") {
+                   this.active_packages = this.active_packages.filter(pkg => pkg !== "cuSpatial/cuProj");
+               }
             },
             imgTypeClickHandler(e, type) {
                 if (this.isDisabled(e.target)) return;
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -756,6 +756,9 @@
             },
             disableUnsupportedPackage(package) {
                 var isDisabled = false;
+                if (this.active_release === "Nightly" && package === "cuSpatial/cuProj") {
+                    isDisabled = true;
+                }
                 return isDisabled;
             },
             isDisabled(e) {
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -317,7 +317,7 @@
                         <div class="option-label"><!-- Second row of packages --></div>
                         <template x-for="package in additional_pip_packages">
                             <div x-on:click="(e) => packagesClickHandler(e, package)"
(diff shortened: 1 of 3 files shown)


_includes/selector.html
```diff
@@ -572,7 +572,15 @@
                 if (this.active_packages.length === 1 && this.active_packages[0] === "Choose Specific Packages") {
                     return "Select at least one package.";
                 } else if (this.active_packages[0] === 'Standard') {
-                    var pkgs = this.additional_pip_packages.flatMap(libraryToPkg);
+                    var pkgs_to_show = this.additional_pip_packages;
+
+                    // Hide cuSpatial and cuProj for Nightly
+                    if (this.active_release === "Nightly") {
+                        pkgs_to_show = pkgs_to_show.filter(pkg => pkg !== "cuSpatial" && pkg !== "cuProj");
+                    }
```
jameslamb, line 580:
```suggestion
                    // possibly filter out any packages that are disabled in specific settings
                    pkgs_to_show = pkgs_to_show.filter(pkg => !disableUnsupportedPackage(pkg))
```

My JavaScript is not the best, but is it possible to do something like this everywhere, instead of repeating the literals `"cuSpatial"` and `"cuProj"` in so many places?

Putting those filters in place would make it easier to do this kind of thing in the future, and reduce the effort to update the selector again when 25.06 becomes the stable version.

jameslamb (approved):
✅ My approval wouldn't count towards a merge here, but I approve. I did click around the preview and it looked to me like everything was working as expected.

For example, seeing `cuSpatial / cuProj` grayed out when choosing "Choose Specific Packages" and RAPIDS 25.06:

image

I left one suggestion for simplifying the code, but treat it as non-blocking... do it only if you want, I know testing these changes and be a bit time-consuming.

_includes/selector.html
jameslamb: blegh ok I see, I'd missed that distinction. Thanks for considering it.

_includes/selector.html
```diff
@@ -644,25 +652,32 @@
...
                 var notes = [];
                 if (this.active_conda_cuda_ver.startsWith("11")) {
                     notes = [...notes, "RAPIDS on CUDA 11 doesn't support <code>channel_priority: strict</code>; use <code>channel_priority: flexible</code> instead"];
                 }
 
+                var pkgs_to_show = this.rapids_meta_pkgs;
+
+                // Hide cuSpatial and cuProj for Nightly
+                if (this.active_release === "Nightly") {
+                    pkgs_to_show = pkgs_to_show.filter(pkg => pkg !== "cuSpatial" && pkg !== "cuProj");
```
raydouglass, line 676: If you select stable, Choose specific packages, and select cuspatial. Then switch to Nightly, the cuspatial button is disabled, but the install script will still include the packages.

image

bdice: Ah, great. I'll try to fix.

bdice: Fixed in af79499. Merging.

PR #603, opened 2025-05-12
feat: add Python 3.13 to selector and make it the new default

gforsyth:
We have Python 3.13 packages now for RAPIDS -- adding those as an option and the new default in the selector.

xref rapidsai/build-planning#120

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -367,7 +367,7 @@
     document.addEventListener('alpine:init', () => {
         Alpine.data('rapids_selector', () => ({
             // default values
-            active_python_ver: "3.13",
+            active_python_ver: "3.12",
             active_conda_cuda_ver: "12",
             active_pip_cuda_ver: "12",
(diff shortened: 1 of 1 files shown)


bdice (changes requested):
We should only use 3.13 for nightlies, not stable packages. That may make the logic massively more complex, unfortunately. Try updating only the nightly version list first and see what happens.

gforsyth:
I'm inclined, then, not to add this to the selector until we release 25.06

bdice:
> I'm inclined, then, not to add this to the selector until we release 25.06

That's fine -- let me give this one quick attempt. If not, we can defer until later.

bdice (approved):
My changes fixed the behavior and it seems right to me now. @gforsyth Please merge if you're happy.

gforsyth:
Ahh, I apparently don't have permissions here.  But merge away @bdice

PR #624, opened 2025-06-18
Add note on miniconda usage

vyasr:
When using miniconda the defaults channel will be included in conda solves by default, which can break any installation of RAPIDS. This PR recommends avoiding the defaults channel when installing RAPIDS.


install/index.md
```diff
@@ -48,6 +48,7 @@ Use the selector tool below to select your preferred method, packages, and envir
 ## Installation Troubleshooting
 
 ### **Conda Issues**
+- Ensure that the `defaults` channel is not included in the conda solve. The `defaults` channel is not compatible with `conda-forge`, which RAPIDS relies on. If you installed conda with [the Miniconda or Anaconda distributions](https://www.anaconda.com/docs/getting-started/miniconda/main#should-i-install-miniconda-or-anaconda-distribution), the `defaults` channel will be included unless you modify your `.condarc` file or specify `nodefaults` in the install commands for RAPIDS packages. If you installed with [Miniforge](https://conda-forge.org/download/) (our recommendation for best compatibility) then the `defaults` channel is not included. Note that in general [mixing `conda-forge` and `defaults` packages is not supported by either](https://conda-forge.org/docs/user/transitioning_from_defaults/).
```
bdice, line 51: The content is good but the formatting needs to fit into the page. The hints below have an info bubble ℹ️ with a problem (e.g. "A `conda create` error occurs") followed by possible solutions.

What are the problems that users will observe if they accidentally mix the channels? You can preview the rendering at https://deploy-preview-624--docs-rapids-ai.netlify.app/install/.

jakirkham: This is reasonable info and agree with Bradley's suggestion

Would also recommend starting this section by linking to our [Conda install docs]( https://docs.rapids.ai/install/#conda ) and pointing out the installer and channel we recommend

Then move to stating `defaults` incompatibilities and how the `defaults` channel might be coming into play

vyasr: Done. Didn't start with our docs since that doesn't really fit the problem->solution structure so well here but I did link them.

install/index.md
```diff
@@ -67,6 +67,11 @@ driver on your machine supports the CUDA version](#system-req) you are trying to
...
 
+<i class="fas fa-info-circle"></i> Even after the above suggestions of updating conda and using `libmamba`/`mamba`, you still see a `conda create error`, or your environment solves but is nonfunctional in some way:<br/>
+Check if any packages in your environment have been installed from the `defaults` channel (you can do that by running `conda list` and inspecting the output). If you installed conda with [the Miniconda or Anaconda distributions](https://www.anaconda.com/docs/getting-started/miniconda/main#should-i-install-miniconda-or-anaconda-distribution), the `defaults` channel will be included unless you modify your `.condarc` file or specify `nodefaults` in the install commands for RAPIDS packages. If you find any packages from `defaults` in your environment, please make those changes and try recreating your environment from scratch. Note that if you installed conda with [Miniforge](https://conda-forge.org/download/) (our recommendation for best compatibility) then the `defaults` channel is not included.
```
bdice: Semantic line breaks. Specifically call out that `defaults` is not supported. Also is it supposed to be `-c nodefaults`?
```suggestion
Check if any packages in your environment have been installed from the `defaults` channel (you can do that by running `conda list` and inspecting the output).
The `defaults` channel is not supported by RAPIDS packages, which are built to be compatible with dependencies from the `conda-forge` channel.
If you installed conda with [the Miniconda or Anaconda distributions](https://www.anaconda.com/docs/getting-started/miniconda/main#should-i-install-miniconda-or-anaconda-distribution), the `defaults` channel will be included unless you modify your `.condarc` file or specify `-c nodefaults` in the install commands for RAPIDS packages.
If you find any packages from `defaults` in your environment, please make those changes and try recreating your environment from scratch.
Note that if you installed conda with [Miniforge](https://conda-forge.org/download/) (our recommendation for best compatibility) then the `defaults` channel is not included.
```

bdice: I guess the line I added about `defaults` not being supported is implied but I want to make that statement earlier so that readers understand why we are telling them to purge `defaults`.

bdice (approved):

PR #631, opened 2025-07-01
pip-wheels-resolved

caryr35:
RAPIDS PIP Wheels pypi issue resolved

### DIFF
--- a/_notices/rsn0049.md
+++ b/_notices/rsn0049.md
@@ -9,8 +9,8 @@ notice_id: 49 # should match notice number
 notice_pin: true # set to true to pin to notice page
 title: "Important Installation Notice : RAPIDS Wheels are missing from pypi.nvidia.com"
 notice_author: RAPIDS Ops
-notice_status: "In Progress"
-notice_status_color: yellow
+notice_status: "Completed"
+notice_status_color: green
 # 'notice_status' and 'notice_status_color' combinations:
 #   "Proposal" - "blue"
 #   "Completed" - "green"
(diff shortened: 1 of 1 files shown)


_notices/rsn0049.md
```diff
@@ -21,11 +21,13 @@ notice_topic: Platform Support Change
...
 # 'notice_updated' should match 'notice_created' until an update is made
-notice_updated: 2025-06-27
+notice_updated: 2025-07-01
 ---
 
 ## Overview
 
+*NVIDIA's pip repo (pypi.nvidia.com) recovery is complete and up and working.  We've confirmed that all of the RAPIDS wheels have been restored to pypi.nvidia.com.*
+
 NVIDIA's pip repo (pypi.nvidia.com) is down, impacting the ability for users to install RAPIDS via pip. The team is working to resolve this as quickly as possible. We apologize for any inconvience this may have caused.
```
jameslamb, line 31:
```suggestion
**2025-07-01:** pypi.nvidia.com recovery is complete. All RAPIDS wheels have been restored to pypi.nvidia.com.

**2025-06-27:** NVIDIA's pip repo (pypi.nvidia.com) is down, impacting the ability for users to install RAPIDS via pip. The team is working to resolve this as quickly as possible. We apologize for any inconvience this may have caused.
```

Because this RSN is more of an "incident report" instead of a single self-contained notice, would you please consider adding dates to the entries?

I've found that really helpful for incidents on status pages, e.g. https://www.githubstatus.com/incidents/ry1gsyjqj4qh

I'd also like to recommend making the update text a bit punchier (for example, we don't need to repeat "NVIDIA's pip repo").

jameslamb (approved):
I left some suggestions, but consider them optional... do what you want with them. Approving so this can be merged without needing another review.

PR #657, opened 2025-08-07
move cuspatial, cuproj docs under 'inactive projects' (attempt 2)

jameslamb:
Fixes https://github.com/rapidsai/ops/issues/4044 (private issue, sorry)

Contributes to https://github.com/rapidsai/build-planning/issues/197

#654 tried to move the `cuproject` and `cuspatial` documentation under a new section at https://docs.rapids.ai/api/ titled "Inactive Projects".

This ended up getting reverted in #658 because it violated some of the assumptions made by post-processing code (e.g. that all libraries except `libucxx` have the same set of versions), resulting in broken links and inaccurate labels.

That led me on a journey to rework that post-processing code to accommodate this:

* #666
* #664
* #663
* #661
* #659

This PR is another attempt to add an "Inactive Projects" section, and I think it'll work this time 😅 

### How I tested this

Checked that the API docs were all in a working state as of the previous deployment: https://github.com/rapidsai/docs/pull/666#issuecomment-3181290663

Tested locally (including fully downloading all of the docs from S3), following this:

https://github.com/rapidsai/docs/blob/b04146f73554cf799cae3143ddc164c9326ae8ed/CONTRIBUTING.md#L37

Everything looked good to me 😁 

Main API docs were formatted correctly, with the expected contents (including for `libucxx` versions!) and all links working:

image

"Inactive Projects" section was formatted correctly, with the expected contents and all links working:

image

Drop-down selectors had the correct set of versions, and links working:

image

Everything at https://deploy-preview-657--docs-rapids-ai.netlify.app/api/ also looks right... but this post-processing doesn't happen on PRs, so the links to project-specifics docs are expected to be broken there.

PR #654, opened 2025-08-06
move cuspatial, cuproj docs under 'inactive projects'

jameslamb:
Fixes https://github.com/rapidsai/ops/issues/4044 (private issue, sorry)

Contributes to https://github.com/rapidsai/build-planning/issues/197

* introduces a new section called "Inactive Projects" at the bottom of https://docs.rapids.ai/api/, to host docs links for projects that are no longer maintained.
* moves `cuproj`, `cusignal`, and `cuspatial` docs links there
  - *with 25.04 as "stable" and 24.02 as "legacy" for `cuproj` / `cuspatial`, as requested in https://github.com/rapidsai/ops/issues/4044*
* updates code for copying docs from S3, to account for these "inactive" projects with old/different versions

### How I tested this

The preview page looks right to me: https://deploy-preview-654--docs-rapids-ai.netlify.app/api/

I see:

* no formatting issues
* correct versions for all the active projects
* correct versions for the inactive projects

We'll have to merge and deploy this to test the S3 changes, but I did a dry run and the output looks correct.

<details><summary>patch to remove AWS API calls (click me)</summary>

```diff
diff --git a/ci/download_from_s3.sh b/ci/download_from_s3.sh
index 4b72289..61abecd 100755
--- a/ci/download_from_s3.sh
+++ b/ci/download_from_s3.sh
@@ -50,17 +50,17 @@ aws_cp() {
   SRC=$1
   DST=$2
 
-  if ! aws s3 ls "${SRC}" > /dev/null; then
-    echo "No files found in ${SRC}. Exiting."
-    exit 1
-  fi
+  # if ! aws s3 ls "${SRC}" > /dev/null; then
+  #   echo "No files found in ${SRC}. Exiting."
+  #   exit 1
+  # fi
 
   echo "Copying ${SRC} to ${DST}"
-  aws s3 cp \
-    --only-show-errors \
-    --recursive \
-    "${SRC}" \
-    "${DST}"
+  # aws s3 cp \
+  #   --only-show-errors \
+  #   --recursive \
+  #   "${SRC}" \
+  #   "${DST}"
 }
 
 # Downloads the RAPIDS libraries' documentation files from S3 and places them
@@ -152,6 +152,6 @@ download_deployment_docs() {
   done
 }
 
-check_dirs
+# check_dirs
 download_lib_docs
 download_deployment_docs
```

</details>

With that patch applied, ran:

```shell
./ci/download_from_s3.sh
```

[...]

PR #661, opened 2025-08-08
post-processing: consolidate scripts, reduce use of global state

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/197 and https://github.com/rapidsai/ops/issues/4044

In #657 (and previously in #654), I'm working on adding an "Inactive Projects" section to https://docs.rapids.ai/api/, and currently struggling to make that work with the post-processing that populates drop-downs like this:

image

ref: https://docs.rapids.ai/api/cudf/stable/user_guide/

To try to make this more manageable to both change and review, this PR proposes some refactoring:

* copying the contents of `ci/customization/util.py` into the only place they're used (to eliminate the indirection of an import)
* switching most functions from relying on global state to instead passing configuration through keyword arguments

This should ONLY be a moving-code-around PR, it isn't intended to change anything about how the docs look or work.

## Notes for Reviewers

I'm sorry, I know this has not been the most fun series of PRs to review. Thanks for your patience.

### How I tested this

Using the steps added to the docs in #659 

```shell
./ci/customization/customize_docs_in_folder.sh "_site/api/cudf"
```

Beyond that, we'll have to merge this to find out if it worked.


ajschmidt8 (approved):
you deserve a :cookie: for wrangling this mess.

jameslamb:
Ha no prob, thanks for reviewing! I'll merge this so we can get some feedback from a deployment.

PR #662, opened 2025-08-08
Update CUDA requirements in 25.08 installation guide

bdice:
In 25.08, there were a few significant changes in CUDA support.

- CUDA 11 was dropped
- CUDA 12.9 support was added
- CUDA 12.8 containers were removed in favor of CUDA 12.9 containers

This PR updates those and also thoroughly revises the installation notes and troubleshooting info. In many places, we had outdated information that is no longer relevant with CUDA 12+.

I also removed notes pertaining to CUDA 11, updated a lot of links, fixed installation instructions for `nvidia-container-toolkit`, and removed workarounds for non-RAPIDS software older than 2023.

PR #663, opened 2025-08-08
post-processing: avoid re-reading config files

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/197 and https://github.com/rapidsai/ops/issues/4044

In #657 (and previously in #654), I'm working on adding an "Inactive Projects" section to https://docs.rapids.ai/api/, and struggling with how long iteration on the API docs takes... both locally and via merges to `main` and waiting for deployments.

After a merge to `main`, it takes slightly over an hour for a deployment to complete, and almost all of that time is spent in "post-processing".

Screenshot 2025-08-08 at 5 12 38 PM

([link to recent build](https://github.com/rapidsai/docs/actions/runs/16837896940/job/47701754745))

This PR attempts to reduce that.

Today, `ci/customization/customize_doc.py` is invoked once per HTML file in all of the API docs. Here in 2025, that's close to 10,000 files.

To make the end-to-end time faster, this proposes the following:

* invoking `customize_doc.py` once and passing it a list of files (instead of `for` looping around it and invoking it once per file)
* only reading static config files like `_data/releases.json` once per post-processing run, instead of once per file

In local testing on my Mac, with these changes `./ci/post-process.sh` run over all of the API docs completes in around 4 minutes. If we get anywhere even close to that amount of a speedup in CI, it'd help with iteration time here AND reduce the lead time to publish updates like #662 for users.

### How I tested this

Using the steps added to the docs in #659 

```shell
./ci/post-process.sh
```

Beyond that, we'll have to merge this to find out if it worked.


jameslamb:
This did not quite work... the post-processing only processed `raft` and then stopped:

```text
Customizing 14721 HTML files
Couldn't identify _site/api/raft/25.10/_static/webpack-macros.html as a supported theme type. Skipping file.
...
--- _site/api/raft/25.10/_static/webpack-macros.html ---
Done customizing
```

([build link](https://github.com/rapidsai/docs/actions/runs/16882230508/job/47820734531#step:7:495))

I think I see the problem... now that the code in `customize_doc.py` is only invoked once (instead of 1 new process per file), using `exit()` like this is no longer appropriate:

https://github.com/rapidsai/docs/blob/0c0433a806349b220851c02c6e6e66506016f6ab/ci/customization/customize_doc.py#L282-L286

I'm not sure why I didn't observe this early stopping when I tested locally, as I did try to run this over the entire set of API docs, for all libraries. Maybe I had other local modifications that led to this not happening.

I'll put up a change shortly, and test it locally in a completely clean environment. Sorry.

PR #669, opened 2025-08-13
post-processing: fix version map, use pre-commit to regenerate it

jameslamb:
Follow-up to fix the issues noted in https://github.com/rapidsai/docs/pull/657#issuecomment-3184788441

In #657, I accidentally removed some projects (like `cucim`, `cuxfilter`, and `rmm`) from https://docs.rapids.ai/api/, via a mix of a bad merge conflict resolution and mistakes made reading the diff.

This fixes that, by restoring the dropped entries in `_data/docs.yml`.

It also proposes changes to prevent this type of thing from slipping through in the future:

* checking `projects-to-versions.json` into source control
* using a `pre-commit` hook to regenerate it
  - *@KyleFromNVIDIA set up something similar for `rapids-medata` and it's been very helpful there* ([code link](https://github.com/rapidsai/rapids-metadata/blob/883506afa46f28406a7b693f81c6a8834c9ed68c/.pre-commit-config.yaml#L54-L69))

### How I tested this

Re-generated the docs site locally, following the docs at https://github.com/rapidsai/docs/blob/b04146f73554cf799cae3143ddc164c9326ae8ed/CONTRIBUTING.md#L37

Confirmed that every expected library was downloaded, the processing worked for all of them, and the API docs pages rendered correctly.

I also tried manually changing entries in `releases.json`, `docs.yml`, and `projects-to-versions.json`, then running

```shell
pre-commit run --all-files
```

In all cases, I saw the hook correctly update that file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,13 @@ repos:
             .devcontainer/devcontainer.json
           )
       - id: check-yaml
+  - repo: local
+    hooks:
+      - id: generate-projects-to-versions
+        name: generate-projects-to-versions
+        entry: ./ci/generate-projects-to-versions.sh
+        language: system
+        pass_filenames: false
   - repo: https://github.com/sirosen/texthooks
     rev: 0.7.1
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,6 +6,9 @@ ci:
   autoupdate_branch: ""
   autoupdate_commit_msg: "[pre-commit.ci] pre-commit autoupdate"
   autoupdate_schedule: quarterly
+  skip:
+    # requires 'jq' and 'yq', which dont' come pre-installed in the pre-commit.ci image
+    - generate-projects-to-versions
   submodules: false
 
 repos:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,7 @@ ci:
   autoupdate_commit_msg: "[pre-commit.ci] pre-commit autoupdate"
   autoupdate_schedule: quarterly
   skip:
-    # requires 'jq' and 'yq', which dont' come pre-installed in the pre-commit.ci image
+    # requires 'jq' and 'yq', which don't come pre-installed in the pre-commit.ci image
     - generate-projects-to-versions
   submodules: false
 
--- a/_data/docs.yml
+++ b/_data/docs.yml
@@ -142,7 +142,7 @@ apis:
     versions:
       # enable or disable links; 0 = disabled, 1 = enabled
       legacy: 0
-      stable: 0
+      stable: 1
       nightly: 1
 
 
--- a/_data/docs.yml
+++ b/_data/docs.yml
@@ -35,6 +35,28 @@ apis:
       legacy: 1
       stable: 1
       nightly: 1
+  cugraph:
+    name: cuGraph
+    path: cugraph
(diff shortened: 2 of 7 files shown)


.pre-commit-config.yaml
```diff
@@ -6,6 +6,9 @@ ci:
   autoupdate_branch: ""
   autoupdate_commit_msg: "[pre-commit.ci] pre-commit autoupdate"
   autoupdate_schedule: quarterly
+  skip:
+    # requires 'jq' and 'yq', which dont' come pre-installed in the pre-commit.ci image
+    - generate-projects-to-versions
```
jameslamb, line 11: This was something I hadn't considered when implementing this in a shell script and then introducing a `pre-commit` hook... for `language: system` types of hooks, you have to manage your own dependencies (https://pre-commit.com/#system).

And on `pre-commit.ci`, we don't have any ability (as far as I can tell) to pre-install things before `pre-commit` runs.

For now, to restore all the API docs, let's just skip this in CI.

I can put up a follow-up switching this to a Python script... that'd be easier to maintain anyway 😁

_data/docs.yml
```diff
@@ -35,6 +35,28 @@ apis:
       legacy: 1
       stable: 1
       nightly: 1
+  cugraph:
+    name: cuGraph
+    path: cugraph
+    desc: 'cuGraph is a GPU accelerated graph analytics library, with functionality like NetworkX, which is seamlessly integrated into the RAPIDS data science platform. cuGraph supports GNNs with PyG, DGL packages, cugraph-service for analytics on a remote graph, and WHOLEGRAPH for memory management.'
```
bdice, line 41: Non-blocking: This needs to be updated, the GNN functionality has been split out.

_data/docs.yml
```diff
@@ -46,6 +68,83 @@ apis:
...
+      nightly: 1
+  rapidsmpf:
+    name: RapidsMPF
+    path: rapidsmpf
+    desc: 'RAPIDS Multi-Process Foundation (rapidsmpf) is a collection of multi-GPU, distributed memory algorithms written in C++ and exposed to Python.'
+    ghlink: https://github.com/rapidsai/rapidsmpf
+    versions:
+      # enable or disable links; 0 = disabled, 1 = enabled
+      legacy: 0
+      stable: 0
```
bdice, line 145: Pretty sure this can be 1 now. Maybe even legacy? Not sure if rapidsmpf was first published in 25.06 or 25.08.
```suggestion
      stable: 1
```

bdice (approved):
Approving with a couple of suggestions

_data/docs.yml
jameslamb: Sure, let's add this.

_data/docs.yml
jameslamb: I'll address this in a follow-up... it does need to be updated (to remove DGL and to mention the specific library names for searchability), but not because the GNN functionality has been split out.

The whole family of `cugraph` packages builds its docs from one repo: https://github.com/rapidsai/cugraph-docs

Instead of uploading from individual projects' CI like the rest of RAPIDS. That is also why you don't see a separate entry for `nx-cugraph` in https://github.com/rapidsai/docs/blob/main/_data/docs.yml, I think.

I'll put up a follow-up with a proposal we can discuss.

cc @acostadon

_data/docs.yml
jameslamb: good way to test the `pre-commit` hook 😁 

https://github.com/rapidsai/docs/pull/669/commits/babe5c0e60797a748ca321e82ddb052bc104ca95

_data/docs.yml
jameslamb: Put up a proposal at https://github.com/rapidsai/docs/pull/671

PR #678, opened 2025-09-10
Fix search

### LINKED ISSUE #677: [BUG] Broken search bar
**Describe the bug**
The search bar in https://github.com/rapidsai/docs is broken. It doesn't perform a search

**To Reproduce**
1. Go to https://github.com/rapidsai/docs
2. Type cudf on the search bar 

If you inspect the site you'll see the error 

Image

**Expected behavior**
Been able to search 

**Desktop (please complete the following information):**
 - Browser  chrome (I tried safari and it's also broken)

bdice:
#642 broke the syntax in `assets/js/just-the-docs.js`. This fixes it.

Closes #677.


assets/js/just-the-docs.js
```diff
@@ -48,19 +48,19 @@ function initNav() {
 
 function initSearch() {
   var request = new XMLHttpRequest();
-  request.open('GET', '{{ "assets/js/search-data.json" | absolute_url }}', true);
+  request.open('GET', '/assets/js/search-data.json', true);
```
bdice, line 51: This change is needed to avoid cross-site requests (which fail) between the testing site (like https://deploy-preview-678--docs-rapids-ai.netlify.app/) and the main site (https://docs.rapids.ai). Otherwise our search data fails to load on the site previews.

assets/js/just-the-docs.js
```diff
@@ -1,9 +1,9 @@
+---
+---
 // SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION & AFFILIATES.
 // All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
----
----
```
bdice, line 6: This is the primary fix needed here. I think the inclusion of the copyright notice comments causes the Jekyll template engine not to trigger. You can see that currently, https://docs.rapids.ai/assets/js/just-the-docs.js appears to have raw template content. Moving the `---` up above the copyright notice comments appears to fix it.

PR #679, opened 2025-09-15
Add UCXX

pentschev:
Add UCXX (Python) docs, depends on https://github.com/rapidsai/ucxx/pull/510.

### DIFF
--- a/_data/docs.yml
+++ b/_data/docs.yml
@@ -144,6 +144,17 @@ apis:
       legacy: 0
       stable: 1
       nightly: 1
+  ucxx:
+    name: UCXX
+    path: ucxx
+    desc: 'UCXX is the Python interface for [UCX(<https://github.com/openucx/ucx), a low-level high-performance networking library. UCX and UCXX support several transport methods including InfiniBand and NVLink while still using traditional networking protocols like TCP.'
+    ghlink: https://github.com/rapidsai/ucxx
+    cllink: https://github.com/rapidsai/ucxx/blob/main/CHANGELOG.md
+    versions:
+      # enable or disable links; 0 = disabled, 1 = enabled
+      legacy: 1
+      stable: 1
+      nightly: 1
 
 
 # RAPIDS "Libs" - lower-level libraries that are building blocks for creating
@@ -197,7 +208,7 @@ libs:
   libucxx:
     name: libucxx
     path: libucxx
-    desc: "UCXX is an object-oriented C++ interface for UCX, with native support for Python bindings."
+    desc: "libucxx is an object-oriented C++ interface for UCX, with native support for Python bindings."
     ghlink: https://github.com/rapidsai/ucxx
     versions:
       # enable or disable links; 0 = disabled, 1 = enabled
--- a/_includes/api-docs.html
+++ b/_includes/api-docs.html
@@ -11,7 +11,7 @@
 #### DOCS {% for version_name in versions -%}
     {%- if api.version-overrides -%}
         **[{{ version_name }} ({{ api.version-overrides[version_name] }})](/api/{{ api.path }}/{{ version_name }})**
-    {%- elsif api.name == "libucxx" -%}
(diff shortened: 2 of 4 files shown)


_data/docs.yml
```diff
@@ -144,6 +144,19 @@ apis:
...
+    name: UCXX
+    path: ucxx
+    desc: 'UCXX is the Python interface for [UCX(<https://github.com/openucx/ucx), a low-level high-performance networking library. UCX and UCXX support several transport methods including InfiniBand and NVLink while still using traditional networking protocols like TCP.'
+    ghlink: https://github.com/rapidsai/ucxx
+    cllink: https://github.com/rapidsai/ucxx/blob/main/CHANGELOG.md
+    versions:
+      # enable or disable links; 0 = disabled, 1 = enabled
+      # TODO: enable 'legacy' once it moves to 0.48, and add version to projects-to-versions.json
+      legacy: 0
+      # TODO: enable 'stable' once it moves to 0.47, and add version to projects-to-versions.json
```
jameslamb, line 157:
```suggestion
      # TODO: enable 'legacy' once it moves to 0.46
      legacy: 0
      # TODO: enable 'stable' once it moves to 0.46
```

I don't think these comments are what you want. `0.46` will be the first version that has docs on S3, so you can enable these other version types once those point to 0.46.

This is configured here:

https://github.com/rapidsai/docs/blob/711b940853d7f91674000e36a3276efd239b7b6b/_data/releases.json#L2-L14

Once `stable` points to `0.46`, it'd be safe to set `stable: 1` for UCXX here.
Once `legacy` points to `0.46`, it'd be safe to set `legacy: 1` for UCXX here.

> _add version to projects-to-versions.json_

No need to document this, this is done automatically by a pre-commit hook that reads from `_data/docs.yml`:

https://github.com/rapidsai/docs/blob/711b940853d7f91674000e36a3276efd239b7b6b/.pre-commit-config.yaml#L38-L46

pentschev: Ah ok, so the comment is relative to `_data/releases.json`, i.e., when `stable`/`legacy` points to `0.46` in `_data/releases.json`, not when we move to a new release. Understood now, thanks for clarifying.

_includes/api-docs.html
```diff
@@ -11,7 +11,7 @@
 #### DOCS {% for version_name in versions -%}
     {%- if api.version-overrides -%}
         **[{{ version_name }} ({{ api.version-overrides[version_name] }})](/api/{{ api.path }}/{{ version_name }})**
-    {%- elsif api.name == "libucxx" -%}
+    {%- elsif api.name == "libucxx" or api.name == "ucxx" -%}
```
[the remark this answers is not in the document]
jameslamb: Thanks for sharing that. I don't have a strong preference either way, my thinking was that `contains` allows us to avoid needing to write `| dropcase` twice, and I couldn't think of any case where this would be wrong (`distributed-ucxx` follows the same versioning scheme as `ucxx`, so this condition would still be correct for that case).

_data/docs.yml
```diff
@@ -144,6 +144,19 @@ apis:
       legacy: 0
       stable: 1
       nightly: 1
+  ucxx:
+    name: UCXX
+    path: ucxx
+    desc: 'UCXX is the Python interface for [UCX(<https://github.com/openucx/ucx), a low-level high-performance networking library. UCX and UCXX support several transport methods including InfiniBand and NVLink while still using traditional networking protocols like TCP.'
```
jameslamb, line 150:
```suggestion
    desc: 'UCXX is the Python interface for [UCX](<https://github.com/openucx/ucx), a low-level high-performance networking library. UCX and UCXX support several transport methods including InfiniBand and NVLink while still using traditional networking protocols like TCP.'
```

Sorry, missed this in my previous reviews.  This link is not rendering correctly.

_includes/api-docs.html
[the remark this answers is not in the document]
pentschev: This is correct now

Screenshot 2025-09-15 at 17 38 11

PR #690, opened 2025-10-01
release-schedule-update adding rapidsmpf

caryr35:
adding rapidsmpf to the cudf release cadence


rockhowse (approved):
pretty simple update.

PR #687, opened 2025-09-23
Add RSN for CUDA major image tags

bdice:
This PR adds an RSN describing the changes in https://github.com/rapidsai/docker/issues/781.

See also: https://github.com/rapidsai/docker/pull/789

PR #692, opened 2025-10-06
Fix default Docker CUDA version

bdice:
This fixes a bug in #687. The default value for `active_docker_cuda_version` should be one of the stable values.


ncclementi (approved):
Thanks @bdice

PR #697, opened 2025-10-15
enable cudf-java stable

AyodeAwe:
`cudf-java` docs have been uploaded to `s3`. This change enables its publication to https://docs.rapids.ai/api


rockhowse (approved):
the current stable release is `25.10` so this change makes sense.

PR #708, opened 2025-11-19
Document what CUDA 13 driver version we 25.10+ supports

robertmaynard:


bdice (approved):
Alternatively, we could say “a supported driver” and point to an external reference of the current driver lifecycle. 525 is EOL and no longer tested.

PR #715, opened 2025-12-10
25.12 - updated Maintainer Docs - Release Planning

rockhowse:
This PR updates the Release Planning section 

* https://docs.rapids.ai/releases/planning/

to bring it up to the current planning/release process post RSN-47 

* https://docs.rapids.ai/notices/rsn0047/

This focuses on using `CALVER (YY.MM.PP)` instead and + an un-changing `main` development branch

xref: https://github.com/rapidsai/build-planning/issues/224

### DIFF
--- a/releases/planning.md
+++ b/releases/planning.md
@@ -26,17 +26,17 @@ Project Leads
 
 ### Timing
 
-**NOTE:** The processes below use the current release of `M.A`, the next release of `M.B` and future release `M.C` (where `B=A+1`, `C=B+1`) for examples.
+**NOTE:** The processes below use the current release of `YY.MA`, the next release of `YY.MB` and future release `YY.MC` (where `MB=MA+2`, `MC=MB+2`) for examples.
 
 Releases are planned in two phases:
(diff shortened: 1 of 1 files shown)


rockhowse:
ok now that we ahve the labels sorted out... I put this up as a draft PR as therei s one key thing I don't know about. 

This page includes a link to a centralized release dashboard of some sort... it originally pointed to a specific repository in `cudf`. 

I added one of the seemingly related "release dashboards" that ops used during the release itself created by @AyodeAwe , but I don't think that's correct either. 

Will need some input to get this set up and aligned with the current release planning paradim and discuss it with @mmccarty  as I believe he has been one of the folks driving the rapids planning.

AyodeAwe:
This looks like very old docs, older than my time here. @bdice do you possibly have context on the page @rockhowse is updating here?

@rockhowse I think you're correct to use one of the release dashboards we've used in recent past. You probably want to involve @caryr35 here as well as he's got a lot of coverage in the area of "release planning".

gforsyth:
Yep, this should get deleted, closing this out

Issue #709, opened 2025-11-20
[DOC] Add platform support matrix


bdice:
We would like to add a version support chart, based on the ideas raised in #466.

We have the following goals:
- Communicate supported versions for the following for each RAPIDS release:
  - CUDA Toolkit/Runtime requirements
  - CUDA Driver requirements
  - Python
  - glibc
  - Compute capabilities (architectures) built/shipped in major RAPIDS products
- Declare the data for the above in a declarative way (YAML, JSON)
  - Current plan is to add platform support information to https://github.com/rapidsai/rapids-metadata and fetch it during docs builds

PR #733, opened 2026-01-09
Add Platform Support page

bdice:
Closes https://github.com/rapidsai/docs/issues/709.

### DIFF
--- a/platform-support/index.md
+++ b/platform-support/index.md
@@ -17,6 +17,8 @@ Note that for the list of supported compute capabilities below, newer GPUs are s
 
 For installation instructions, see the [Installation Guide](/install/).
 
+**Releases:** {% for release in site.data.platform_support.releases %}[{{ release.version }}](#rapids-{{ release.version | replace: ".", "" }}){% unless forloop.last %}, {% endunless %}{% endfor %}
+
 {% for release in site.data.platform_support.releases %}
 ---
 


rockhowse (approved):
While I don't have the requisite codeowners to officially approve, I am going to add my approval here as this is something end users like myself rely on when reviewing solutions if we get ourselves into trouble by trying to just use `latest` of everything and failing 😬 

The structure makes sense and accurately captures the major sets of dependencies required to keep in mind when looking to use a specific version of `RAPIDS` and/or planning for future upgrades.

platform-support/index.md
```diff
@@ -0,0 +1,56 @@
...
+{: .fs-5 }
+
+| | {% for cuda in release.cuda %}CUDA {{ cuda.major }}{% unless forloop.last %} | {% endunless %}{% endfor %} |
+|:--|{% for cuda in release.cuda %}:--|{% endfor %}
+| **Toolkit** | {% for cuda in release.cuda %}{{ cuda.toolkit_min }}{% if cuda.toolkit_min != cuda.toolkit_max %} - {{ cuda.toolkit_max }}{% endif %}{% unless forloop.last %} | {% endunless %}{% endfor %} |
+| **Driver** | {% for cuda in release.cuda %}{{ cuda.driver_min }}+{% unless forloop.last %} | {% endunless %}{% endfor %} |
+| **Compute Capability** | {% for cuda in release.cuda %}{% for cc in cuda.compute_capability %}{{ cc.name }} ({% if cc.sm.first %}{{ cc.sm | join: ", " }}{% else %}{{ cc.sm }}{% endif %}){% unless forloop.last %}, {% endunless %}{% endfor %} or newer{% unless forloop.last %} | {% endunless %}{% endfor %} |
+
+#### <i class="fas fa-hammer"></i> Source Builds
+{: .fs-5 }
```
jameslamb: I appreciate not wanting to get too carried away with this "Source Builds" table, but would you consider adding host compiler(s) / version(s) here?

I'd imagine that's one of the first places to look when someone tries to build a RAPIDS project from source outside of our infrastructure and hits issues.

And it is probably worth noting that support for compiling with `clang` is "best-effort / unofficial" (I think that's accurate in 2026, though not 100% sure).

jameslamb (approved):
Everything included here looks accurate to me.

I left a few comments that are all suggestions for expanding what's covered... go as far as you want with those, including not changing anything at all.

Thanks for doing this! Super super helpful to have this in documentation we can link to, I hope we could use it to eliminate some duplication across docs in the individual library repos.

PR #735, opened 2026-01-13
add CUDA 13.1 support to install selector

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/236

Adds CUDA 13.1 to the release selector for nightly packages. RAPIDS 26.02 nightlies almost all now support CUDA 13.1 (just cuGraph is left, see https://github.com/rapidsai/cugraph/pull/5383).

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -449,7 +449,7 @@
                     },
                     "Nightly": {
                         "12": ["12.2", "12.9"],
-                        "13": ["13.0"]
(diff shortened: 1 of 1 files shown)


jameslamb:
The preview webhooks seem not to be working, these have showed as "in progress" for 18+ hours:

image

But the URLs are predictable, so you can see the preview: https://deploy-preview-735--docs-rapids-ai.netlify.app/

And it looks ok there.

image

PR #768, opened 2026-04-09
Update release versions for 26.04

AyodeAwe:
Roll version labels for 26.04 release:
- legacy: 25.12 -> 26.02
- stable: 26.02 -> 26.04
- nightly: 26.04 -> 26.06
- next_nightly: 26.06 -> 26.08

Also enables stable docs for nvforest (first release).

### DIFF
--- a/_data/platform_support.yml
+++ b/_data/platform_support.yml
@@ -1,4 +1,56 @@
 releases:
+  - version: "26.04"
+    python:
+      - "3.11"
+      - "3.12"
+      - "3.13"
+      - "3.14"
+    glibc_min: "2.28"
+    cpu_arch:
+      - "x86_64"
+      - "aarch64"
+    os_support:
+      - "Ubuntu 22.04"
+      - "Ubuntu 24.04"
+      - "Rocky Linux 8"
+    source_build:
+      cccl: "3.2.0"
+      nvcomp: "5.1.0.21"
+      gcc: "13.3+"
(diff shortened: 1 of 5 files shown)


jameslamb (approved):
Looks great, thanks for catching adding nvForest `stable` docs and for updating the platform support.

I think we should make the default Python version in the install selector 3.14:

https://github.com/rapidsai/docs/blob/25e8185219ac2cf38ad997006ce8b15681cf43ca/_includes/selector.html#L373

That's generally helpful to keep nudging folks towards newer versions. AFAIK the only RAPIDS packages that don't have 3.14 support yet are those from `cugraph-gnn` (https://github.com/rapidsai/build-infra/issues/348) and those aren't mentioned in the install selector.

_data/platform_support.yml
```diff
@@ -1,4 +1,56 @@
...
+    glibc_min: "2.28"
+    cpu_arch:
+      - "x86_64"
+      - "aarch64"
+    os_support:
+      - "Ubuntu 22.04"
+      - "Ubuntu 24.04"
+      - "Rocky Linux 8"
+    source_build:
+      cccl: "3.2.0"
```
bdice: Source: https://github.com/rapidsai/rapids-cmake/blob/release/26.04/rapids-cmake/cpm/versions.json#L16
```suggestion
      cccl: "3.3.0"
```

PR #769, opened 2026-04-10
Remove special-casing of nvForest installation instructions.

csadorf:
Previously gated on nightlies since there was no stable release yet.

Follow-up to https://github.com/rapidsai/docs/pull/754


bdice (approved):

csadorf:
I realize that the current docs still show 26.02 as stable. Maybe we have to wait for that to be updated first?

PR #782, opened 2026-05-21
docs(git): delete out-of-date git page

gforsyth:
Again, hopelessly out-of-date, also written to a mixed audience of people who have

1. never used git before
2. seem to know what git LFS is?

Resolves #721


jameslamb (approved):
I support this, these are wayyy out of date which suggests that probably no one has checked them and reported issues.

For the stuff about file size, we have a CI check called `rapids-size-check` that runs in every RAPIDS repo: https://github.com/rapidsai/shared-workflows/blob/ece43bbc9347340721e5cc576e6ba69148176d3f/.github/workflows/checks.yaml#L70

PR #783, opened 2026-05-21
docs(style): delete style doc

gforsyth:
Out of date, and otherwise specifies things that don't need to be said.


jameslamb (approved):
Agree! Projects have `pre-commit` and other tools for this, we don't need a central RAPIDS doc for it.

PR #784, opened 2026-05-21
docs(testing): remove testing doc

gforsyth:
A lot of the information on this page is out-of-date.  The useful information is on pages that are linked elsewhere.

I'll also posit that for the "persona" of rapids maintainer, this information is redundant and over-simplified.

A document outlining some of the _incredibly complicated_ testing that goes into some of the zero-code-change APIs and how to inherit upstream test suites would be interesting and useful.


jameslamb (approved):
These are pretty out of date and at this point the details vary widely across libraries. I agree this page can just go.

PR #785, opened 2026-05-21
docs(triage): remove triage doc

gforsyth:
This hasn't been touched since 2019 -- all the links are rotted out.

And I believe this dates from an era with more centralized planning process.  

This information is either out-of-date, or not necessary to document on a page for active devs and project leads.


jameslamb (approved):
Agree, thanks for cleaning this up

PR #788, opened 2026-05-27
Add SECURITY.md

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/281

* adds a `SECURITY.md` describing how to report security vulnerabilities

### Why not just set this org-wide?

An org-wide default is set at https://github.com/rapidsai/.github/blob/main/SECURITY.md, but adding an actual file in each repo offers a few benefits:

* ensures security policy travels with the repo to forks, clones, mirrors, etc.
* allows per-repo governance over the security policy (via PR review, CODEOWNERS, etc.)

### DIFF
--- a/ci/customization/customize_doc.py
+++ b/ci/customization/customize_doc.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023-2025, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION & AFFILIATES.
 # All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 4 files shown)


ci/customization/customize_doc.py
```diff
@@ -1,18 +1,19 @@
...
-import re
-import sys
 import json
 import os
-from bs4 import BeautifulSoup
+import re
+import sys
 from copy import deepcopy
 
+from bs4 import BeautifulSoup
```
jameslamb, line 15: This change was made by `pre-commit`. I don't really understand how `main` got into this state, I double-checked and `pre-commit` checks are enforced in CI here.

jameslamb: image

gforsyth (approved):

PR #791, opened 2026-06-11
docs(pr-artifacts): update usage examples of `rapids-get-pr-artifact`

gforsyth:
Updates usage examples for `rapids-get-pr-artifact` to reflect the changes made in rapidsai/gha-tools#262 as a part of the artifact name standardization work in rapidsai/build-planning#270


gforsyth:
The `dask-cuda` example doesn't reflect upstream YET because I haven't made those changes, but that is the future state (and I'll open the PR to make it the future state in a few minutes)

bdice (approved):

PR #790, opened 2026-06-05
Update release versions for 26.06

jolorunyomi:
Roll version labels for 26.06 release:

legacy: 26.02 -> 26.04
stable: 26.04 -> 26.06
nightly: 26.06 -> 26.08
next_nightly: 26.08 -> 26.10

PR #758, opened 2026-03-18
Remove staggered dev/burndown schedule from release docs

AyodeAwe:
Since 25.12, all RAPIDS repos enter burndown simultaneously -- release branches are created for all repos at the same time. The docs still show a two-phase staggered schedule for development and burndown (cuDF group vs "others"), which has caused confusion -- most recently cucim PR #946 going to `main` instead of `release/26.04` because the docs said cucim wasn't in burndown yet.

This PR collapses the dev and burndown phases into single entries while keeping code freeze staggered (cuDF group freezes before others).

PR #798, opened 2026-06-26
fix(dates): burndown starts on the same day everywhere

gforsyth:
We updated the docs to indicate that the start of burndown is no longer staggered in #758 -- those changes were accidentally reverted as part of #790.

I've corrected the dates -- the `days` variable here is _way_ too manual and inconsistent, so no promises.

I _think_ it's inclusive of start- and end-date, but doesn't count weekends, holidays, or NVIDIA free days, but there are some existing counts that don't include the end-date.

That can be rectified later (or we can remove the "days" column).

### DIFF
--- a/_data/releases.json
+++ b/_data/releases.json
@@ -23,9 +23,9 @@
       "days": "6"
     },
     "other_burndown": {
-      "start": "Jul 23 2026",
+      "start": "Jul 16 2026",
       "end": "Jul 29 2026",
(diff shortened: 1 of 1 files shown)


jameslamb (approved):

jameslamb (approved):
The dates are more important than the "days". I don't know the right interpretation either but yours sounds good to me.

PR #800, opened 2026-06-29
Add NVIDIA docs selector styling

bdice:
## Summary
- Add a NVIDIA-themed selector stylesheet and use it when rendered docs load NVIDIA Sphinx Theme.
- Keep RAPIDS custom.css for non-NVIDIA pages, including stable docs that should remain RAPIDS purple.
- Fix selector icons without relying on Font Awesome pseudo-glyphs.
- Filter selector project entries to exclude inactive, hidden, and legacy-only projects.

## Validation
- Pre-commit hooks passed while creating each commit.
- python -m py_compile ci/customization/customize_doc.py
- Verified selector metadata includes active projects such as cudf, cuvs, and rmm, while excluding inactive/hidden projects such as cuproj, cusignal, cuspatial, libcuproj, libcuspatial, and librmm.
- Rendered local docs for RMM nightly/stable plus cuVS/cuDF pages and verified NVIDIA-themed nightly docs use custom_nvidia.css while RMM stable retains RAPIDS custom.css.

### DIFF
--- a/assets/css/custom.css
+++ b/assets/css/custom.css
@@ -438,7 +438,7 @@ legend {
 
 .rapids-selector__selected {
   cursor: pointer;
-  padding: 0.5em 0;
+  padding: 0.5em 2em 0.5em 0;
 }
 
 .rapids-selector__selected::after {
@@ -463,7 +463,7 @@ legend {
 #rapids-jtd-container .rapids-selector__selected::after,
 #rapids-pydata-container .rapids-selector__selected::after {
   position: absolute;
-  right: 0;
+  right: 0.75em;
 }
 
 .rapids-selector__menu {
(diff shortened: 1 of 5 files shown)


assets/css/custom.css
```diff
@@ -506,6 +506,39 @@ a.rapids-selector__menu-item--selected {
...
+  position: absolute;
+  top: 0.85em;
+  right: 0.75em;
+  transform: rotate(225deg);
+  transform-origin: 65% 65%;
+}
+
+#rapids-pydata-container .rapids-selector--hidden .rapids-selector__selected::after {
+  transform: rotate(45deg);
+}
```
bdice, line 540: These changes fix a problem with icon rendering that looks like this:

## Before (icons are broken)

Image

## After (embedded icons)

Image

ci/customization/customize_doc.py
```diff
@@ -51,6 +54,26 @@ def get_lib_from_fp(*, filepath: str, lib_path_dict: dict) -> str:
...
+    for docs_key in ["apis", "libs"]:
+        for project_name, project_details in docs_config[docs_key].items():
+            versions = project_details["versions"]
+            if project_details.get("hidden", False):
+                continue
+            if not (versions.get("stable") == 1 or versions.get("nightly") == 1):
+                continue
+            selector_projects.add(project_name)
+
+    return selector_projects
```
bdice, line 74: Previously we had projects like `cuspatial` and `libcuproj` still showing up in our docs project selector. Inactive projects should be filtered out here.

## Before (some inactive projects)

Image

## After (only active projects)

Image

assets/css/custom_nvidia.css
bdice, line 1: This is a new stylesheet for projects that are using the NVIDIA Sphinx Theme instead of the RAPIDS-customized PyData Sphinx Theme. We inject custom CSS/JS for all projects, so we should make it match the right style for the project's chosen theme.

## Before (mix of green and purple)

Image

## After (all green ... and uses the NVIDIA theme's default fonts which are better! 💚)

Image

assets/css/custom.css
jameslamb: How did you generate these screenshots? Did you run all of the post-processing locally?

The API docs aren't visible in the preview (#6800.

ci/customization/customize_doc.py
jameslamb: Where is this drop-down? I can't find a page on https://deploy-preview-800--docs-rapids-ai.netlify.app/ that has it.

jameslamb (approved):
Because these changes seem to only focus on the API docs (the other pages in the preview are all still showing the RAPIDS bright purple, not NVIDIA green), the only ways I know to test it are:

* run the S3-downloading and post-processing manually (manual, difficult, takes more than 30 minutes)
* merge this and see what happens

If you're confident in these changes, fine with me to merge them.

I'm also useless as a CSS reviewer, I have 0 knowledge of that.

bdice:
> If you're confident in these changes, fine with me to merge them.

Yes, I did extensive local testing. My agent found a nice way to pull the S3 artifacts and I am happy with the results -- which I used for the screenshots, as proof that it "actually works."

If this goes awry, we can always fix it up or revert.

PR #803, opened 2026-06-30
Add RSN for CuPy 14 and NumPy 2

### LINKED ISSUE #279: feat(deps): raise `cupy` lower bound to `>=14`
The current lower bound for `cupy` across RAPIDS projects is `13.6.0`.  What is the appetite for us bumping that lower-bound to `>=14`?

In `14.0.0`, `cupy` added a `ctk` install extra, that handles installing all the required `cuda-toolkit` packages from PyPI.  If we bump the lower-bound, it becomes (more) possible for users to have a single install command that leaves them with a fully-formed installation (not missing required `cuda-toolkit` shared objects).

For wheel-based ci tests, raising the lower-bound of `cupy` would also allow us to swap the base image for our `citestwheel` from `cuda-devel-*` to `cuda-base-*`.  We currently use the `devel` as the base image because we need to bundle the shared objects that can now be installed using the `ctk` extra in `cupy`.

If we can swap the base image, it would help with CI spin-up times -- currently, the `citestwheel` images are on the order of around 11GB.  Switching to `base` reduces the image size to ~420MB . This saving in image size IS slightly offset by the size of the `ctk` extra downloads, but it is still a smaller amount of data to transfer, and not every RAPIDS project will end up pulling `cupy[ctk]`.

The downsides I can think of are:

1. For users who already have a local CTK installed, if `cupy-cuda12x[ctk]` is a dependency of, say, `cudf`, they'll download an additional copy of some pretty hefty libraries.
2. If there are users who are constrained to `cupy<14` for some reason, they won't be able to use newer versions of RAPIDS 

I believe the current projects with _direct_ `cupy` _runtime_ dependencies (so the projects that might incur the "extra CTK download" penalty) are:

- `cucim`
- `cudf`
- `cugraph-pyg`
- `cugraph`
- `cuml`
- `cuopt`
- `cuxfilter`
- `kvikio`
- `nvforest`
- `nx-cugraph`
- `rapidsmpf`

Libraries that depend directly on `cupy` for tests:
- `cuvs`
- `raft`
- `rmm`
- `ucxx`

bdice:
Adds RSN 61 for the RAPIDS 26.08 CuPy 14.0.1+ and NumPy 2 requirement, based on https://github.com/rapidsai/build-planning/issues/279#issuecomment-4783280020.

The notice keeps the recommended title, overview, and impact from the planning comment, with RAPIDS docs notice metadata for an in-progress v26.08+ platform support change.

Closes https://github.com/rapidsai/build-planning/issues/279.

### DIFF
--- a/_notices/rsn0061.md
+++ b/_notices/rsn0061.md
@@ -29,15 +29,14 @@ notice_updated: 2026-06-30
 
 RAPIDS will require CuPy version 14.0.1 or newer in the 26.08 release.
 
+As a consequence, RAPIDS will require NumPy 2 as CuPy 14 requires it.
+
 This simplifies RAPIDS installation, because CuPy 14 can now be used without
 installing a system-wide CUDA Toolkit. Users can rely on
 [CUDA components provided via PyPI](https://pypi.org/project/cuda-toolkit/).
 Only the CUDA driver needs to be pre-installed. See the
 [CuPy 14 release blog](https://medium.com/cupy-team/announcing-cupy-v14-e8515ec05fca)
 for more information.
-
-This requirement also means that RAPIDS will require NumPy 2, because of CuPy
-14's requirements.
 
 ## Impact
 
--- a/_notices/rsn0061.md
+++ b/_notices/rsn0061.md
@@ -1,0 +1,45 @@
+---
+layout: notice
+parent: RAPIDS Support Notices
+grand_parent: RAPIDS Notices
+nav_exclude: true
+notice_type: rsn
+# Update meta-data for notice
+notice_id: 61 # should match notice number
+notice_pin: true # set to true to pin to notice page
+
+title: "RAPIDS 26.08 will require CuPy 14 and NumPy 2"
+notice_author: RAPIDS TPM
+notice_status: In Progress
+notice_status_color: yellow
+# 'notice_status' and 'notice_status_color' combinations:
+#   "Proposal" - "blue"
+#   "Completed" - "green"
+#   "Review" - "purple"
+#   "In Progress" - "yellow"
+#   "Closed" - "red"
+notice_topic: Platform Support Change
+notice_rapids_version: "v26.08+"
+notice_created: 2026-06-30
+# 'notice_updated' should match 'notice_created' until an update is made
+notice_updated: 2026-06-30
+---
+
+## Overview
+
+RAPIDS will require CuPy version 14.0.1 or newer in the 26.08 release.
+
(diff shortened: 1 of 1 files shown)


gforsyth (approved):

jameslamb (approved):

_notices/rsn0061.md
```diff
@@ -0,0 +1,45 @@
...
+
+This simplifies RAPIDS installation, because CuPy 14 can now be used without
+installing a system-wide CUDA Toolkit. Users can rely on
+[CUDA components provided via PyPI](https://pypi.org/project/cuda-toolkit/).
+Only the CUDA driver needs to be pre-installed. See the
+[CuPy 14 release blog](https://medium.com/cupy-team/announcing-cupy-v14-e8515ec05fca)
+for more information.
+
+This requirement also means that RAPIDS will require NumPy 2, because of CuPy
+14's requirements.
```
jakirkham: Would move this to the top and maybe shorten

_notices/rsn0061.md
```diff
@@ -0,0 +1,45 @@
...
+notice_topic: Platform Support Change
+notice_rapids_version: "v26.08+"
+notice_created: 2026-06-30
+# 'notice_updated' should match 'notice_created' until an update is made
+notice_updated: 2026-06-30
+---
+
+## Overview
+
+RAPIDS will require CuPy version 14.0.1 or newer in the 26.08 release.
```
jakirkham, line 30: Perhaps something like this?

```suggestion
RAPIDS will require CuPy version 14.0.1 or newer in the 26.08 release.

As a consequence, RAPIDS will require NumPy 2 as CuPy 14 requires it.
```

jakirkham (commented):
Thanks Bradley! 🙏

Looks good

Would suggest pulling the NumPy 2 line to the top of the description to make it easier to see

Details can then follow after

jameslamb (approved):

_notices/rsn0061.md
jakirkham: Addressed in commit: [`b2528fb`]( https://github.com/rapidsai/docs/pull/803/changes/b2528fba0d5ccf47bebe091a374f2762bdf37ee4 )

_notices/rsn0061.md
jakirkham: Addressed in commit: [`b2528fb`]( https://github.com/rapidsai/docs/pull/803/changes/b2528fba0d5ccf47bebe091a374f2762bdf37ee4 )

jakirkham (approved):

PR #809, opened 2026-07-07
move cuxfilter to 'inactive' projects, remove references

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/291

* moves `cuxfilter` to the "inactive projects' section of the API docs
* removes references encouraging its use

### DIFF
--- a/_data/docs.yml
+++ b/_data/docs.yml
@@ -46,17 +46,6 @@ apis:
       legacy: 1
       stable: 1
       nightly: 1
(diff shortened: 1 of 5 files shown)


jameslamb:
Preview looks fine to me, `cuxfilter` correctly moved to the "Inactive Projects" section and the links all look right.

image

bdice (approved):

PR #810, opened 2026-07-07
Add RSN for cuml.tsa deprecation

csadorf:
Adds RSN 62 for the `cuml.tsa` deprecation, including the `26.08` deprecation notice, affected APIs, `26.12` removal timeline, and migration guidance.

Part of https://github.com/rapidsai/cuml/issues/8343

### DIFF
--- a/_notices/rsn0062.md
+++ b/_notices/rsn0062.md
@@ -1,0 +1,67 @@
+---
+layout: notice
+parent: RAPIDS Support Notices
+grand_parent: RAPIDS Notices
+nav_exclude: true
+notice_type: rsn
+# Update meta-data for notice
+notice_id: 62 # should match notice number
+notice_pin: true # set to true to pin to notice page
+
+title: "Deprecation and planned removal of cuml.tsa"
+notice_author: RAPIDS TPM
+notice_status: In Progress
+notice_status_color: yellow
+# 'notice_status' and 'notice_status_color' combinations:
+#   "Proposal" - "blue"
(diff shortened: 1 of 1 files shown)


bdice (changes requested):
Why does this warrant an RSN? Is library documentation insufficient?

Typically RSNs that we write for deprecations are for changes in platform support (like dropping Python or CUDA versions) or entire libraries/packages that are inactive (like cuspatial or cuxfilter). For API deprecation I recommend advertising that in the library docs and with an appropriate `FutureWarning`, especially because no replacement is offered. Only cuML is affected (not all of RAPIDS) and it seems that an RSN is not necessary to meet the acceptance criteria in the linked issue.

csadorf:
> Why does this warrant an RSN? Is library documentation insufficient?
> 
> Typically RSNs that we write for deprecations are for changes in platform support (like dropping Python or CUDA versions) or entire libraries/packages that are inactive (like cuspatial or cuxfilter). For API deprecation I recommend advertising that in the library docs and with an appropriate `FutureWarning`, especially because no replacement is offered. Only cuML is affected (not all of RAPIDS) and it seems that an RSN is not necessary to meet the acceptance criteria in the linked issue.

@mgrauer @btepera FYI, I agree with @bdice 's assessment here. I will withdraw the proposed RSN.

bdice:
Thanks. Just want to avoid bloat / drift from the purpose of RAPIDS-wide notices. :)

PR #811, opened 2026-07-08
Remove TensorFlow and enable PyTorch with CUDA 13

bdice:
## Summary

- Removes TensorFlow from the install selector's **Additional Packages** options and deletes obsolete compatibility notes.
- Allows PyTorch to be selected with CUDA 13. I validated that current official PyTorch wheels and conda-forge packages support CUDA 13.


_includes/selector.html
```diff
@@ -761,9 +754,6 @@
             },
             disableUnsupportedCuda(cuda_version) {
                 var isDisabled = false;
-                // PyTorch and TensorFlow do not support CUDA 13 yet
-                if (this.active_additional_packages.includes("PyTorch") && (!cuda_version.startsWith("12"))) isDisabled = true;
```
jameslamb, line 765: ouch, this definitely should have been caught before, PyTorch has had CUDA 13 support for a while tnow. Thanks for removing it.

jameslamb (approved):
I support this, thanks.

PR #815, opened 2026-07-27
Require successful PR builds

bdice:
Require all PR jobs and exclude imported API docs from portal-specific checks.


jameslamb (approved):
Great, thank you. I've added `pr-builder` to the branch protection on `main`.

Image

PR #816, opened 2026-07-27
Remove dask-cuda from package selectors

### LINKED ISSUE #813: Remove dask-cuda from specific package selectors
Discussed with @ncclementi -- we see `dask-cuda` is included in `cudf` and `cuml` specific package selectors, only for nightlies. This is probably a bug, we don't think it is needed.

bdice:
Fixes #813.

Remove the nightly-only dask-cuda package injection from the selector output. I don't know why this was added originally, or if there's a reason it might still be needed, but I can't think of any reasons. The git blame traces back to https://github.com/rapidsai/docs/pull/450.

### DIFF
--- a/_includes/selector.html
+++ b/_includes/selector.html
@@ -668,11 +668,6 @@
                         .flatMap(libraryToPkg);
                 }
 
-                // pkgs.length == 2 because it includes the "Choose Specific Packages" option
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
Totally agree with removing this, I'm not sure of the origin story either but it feels out of place today.

PR #825, opened 2026-08-13
Fix v26.10 release schedule dates

bdice:
## Summary

Update the v26.10 burn-down and code-freeze dates and durations to the current tentative plan.

The release schedule page will show the corrected timeline for contributors planning the v26.10 release.


jameslamb (approved):
thank you

PR #827, opened 2026-08-20
Migrate the RAPIDS docs portal while preserving API hosting

bdice:
## Summary

Move the RAPIDS documentation portal to https://docs.nvidia.com/datascience/ while keeping unmigrated API documentation available on `docs.rapids.ai`.

- Publish the portal with `rapidsai/shared-actions/publish-docs`. Exclude `deployment/*` from uploads and deletions so `rapidsai/deployment` retains ownership of that subtree.
- Keep a separate Netlify deployment for unmigrated libraries and releases. Generate permanent redirects for portal pages and migrated API versions, without a global catch-all that would intercept hosted API content or shared assets.
- Update portal links, canonical URLs, and the notices RSS feed for the new location. Resolve API links using the existing library migration metadata.

## Rollout checks

- Confirm the NVIDIA publishing secrets and existing source-S3/Netlify configuration, including a populated Akamai notification email list.
- Verify the first NVIDIA publish and Akamai cache refresh, then check redirects and remaining API content on `docs.rapids.ai`.

### DIFF
--- a/.github/workflows/deploy-redirects.yaml
+++ b/.github/workflows/deploy-redirects.yaml
@@ -4,14 +4,6 @@ on:
   schedule:
     - cron: "0 9 * * *"
   workflow_dispatch:
-    inputs:
-      status:
-        description: HTTP status for generated redirects
-        type: choice
-        options:
-          - "302"
-          - "301"
-        default: "302"
   push:
     branches:
       - main
@@ -55,12 +47,9 @@ jobs:
       - name: Assemble and validate API documentation
         run: make assemble
       - name: Generate compatibility redirects
-        env:
-          REDIRECT_STATUS: ${{ inputs.status || vars.DOCS_RAPIDSAI_REDIRECT_STATUS || '302' }}
         run: |
           uv run python scripts/generate_redirect_site.py \
-            --output _site/_redirects \
-            --status "${REDIRECT_STATUS}"
+            --output _site/_redirects
       - name: Deploy compatibility site to docs.rapids.ai
         env:
           NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_API_TOKEN }}
--- a/.github/workflows/deploy-redirects.yaml
+++ b/.github/workflows/deploy-redirects.yaml
@@ -1,0 +1,53 @@
+name: Deploy docs.rapids.ai redirects
+
+on:
+  workflow_dispatch:
+    inputs:
+      status:
+        description: HTTP status for generated redirects
+        type: choice
+        options:
+          - "302"
+          - "301"
+        default: "302"
+
+concurrency:
+  group: deploy-docs-rapids-ai-redirects
+  cancel-in-progress: false
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  contents: read
+
+jobs:
+  deploy:
+    name: Generate and deploy redirect-only site
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - name: Set up uv
+        uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39 # v8.2.0
+        with:
+          enable-cache: true
+      - name: Install dependencies
+        run: uv sync --locked
+      - name: Generate redirect-only site
+        env:
+          REDIRECT_STATUS: ${{ inputs.status }}
+        run: |
+          uv run python scripts/generate_redirect_site.py \
+            --output _redirect_site/_redirects \
+            --status "${REDIRECT_STATUS}" \
+            --require-complete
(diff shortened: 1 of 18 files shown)


.github/workflows/deploy-redirects.yaml
```diff
@@ -0,0 +1,53 @@
+name: Deploy docs.rapids.ai redirects
+
+on:
+  workflow_dispatch:
+    inputs:
+      status:
```
bdice: Remove this option, make it always 301.

.github/workflows/pr.yaml
```diff
@@ -55,11 +53,8 @@ jobs:
...
-          role-to-assume: ${{ vars.AWS_ROLE_ARN }}
-          aws-region: ${{ vars.AWS_REGION }}
-          role-duration-seconds: 7200
-      - name: Assemble and validate complete documentation site
-        run: make assemble
+      - name: Generate redirect preview
+        run: |
+          uv run python scripts/generate_redirect_site.py \
+            --output _redirect_site/_redirects \
+            --status 302
```
bdice: Remove this option, always use status 301.

extensions/rapids_docs/lifecycle.py
```diff
@@ -27,6 +27,8 @@ def _jinja_environment(app) -> Environment:
 
 def _context(app, docname: str = "index") -> dict:
     data = app.rapids_portal_data
+    config = getattr(app, "config", None)
+    site_baseurl = getattr(config, "html_baseurl", "https://docs.rapids.ai/")
```
bdice: The default should be https://docs.nvidia.com/datascience/.

jayavenkatesh19:
I just went through this PR, and I think we can make a small change in the `publish-docs` shared action to make the workflow better. 

This workflow publishes at the `/datascience` root, so it has one requirement the current action does not support, which is to ensure that `/datascience/deployment/*` is excluded from both upload and delete, since that subtree is published by `rapidsai/deployment`

One idea to resolve this would be to add an optional `target-s3-exclude` input to `publish-docs`. This workflow would then look something like this:
```
uses: rapidsai/shared-actions/publish-docs@main
with:
  source-path: _site
  target-s3-key: datascience
  target-s3-exclude: deployment/*
  # existing NVIDIA AWS and Akamai inputs
``` 

That would eliminate the need for AWS setup, S3 sync, Akamai dependency installation, flush requests and cleanup which is all done in inline code in the workflow, in favour of using the `shared-action`, while maintaining independence of the deployment docs.

Happy to open a PR on `shared-actions`, if we decide to go with this.

.github/workflows/deploy.yaml
```diff
@@ -32,26 +33,29 @@ jobs:
...
+          persist-credentials: false
+          sparse-checkout: tools
+      - name: Add gha-tools to PATH
+        run: echo "${GITHUB_WORKSPACE}/gha-tools/tools" >> "${GITHUB_PATH}"
+      - name: Publish portal to NVIDIA docs
+        uses: rapidsai/shared-actions/publish-docs@82e2c50e4703a224de0bc8d0f6e5d12dcab68db7
+        with:
+          source-path: _site
+          target-s3-key: datascience
+          target-s3-exclude: deployment/*
```
josephine-wolf-oberholtzer, line 64: Nice!

.github/workflows/pr.yaml
```diff
@@ -55,11 +57,15 @@ jobs:
...
           role-to-assume: ${{ vars.AWS_ROLE_ARN }}
           aws-region: ${{ vars.AWS_REGION }}
           role-duration-seconds: 7200
-      - name: Assemble and validate complete documentation site
+      - name: Assemble and validate API documentation
         run: make assemble
+      - name: Generate compatibility redirects
+        run: |
+          uv run python scripts/generate_redirect_site.py \
+            --output _site/_redirects
```
josephine-wolf-oberholtzer: Duplicate `Publish portal to NVIDIA docs` (and related precursor steps) from `deploy.yaml` at the end of this workflow, and set `dry-run: true`. That will perform a dry run of the S3 upload and validate all the flags are sane.

josephine-wolf-oberholtzer (changes requested):

jayavenkatesh19:
These are the things I added in my commit focused mainly on workflow and analytics changes:

- Google analytics comes from the theme via `public_docs_features`, mirroring `cudf` and `deployment`. 
- `deploy.yaml` only publishes from main, gated with a new env variable called `UPLOAD`. This ensures that when `workflow_dispatch` is run on a different branch, the build happens but the upload is skipped. 
- The compat site deploy is now within the same workflow as a separate job, ensuring that it runs only after the publish to `docs.nvidia.com` succeeds. Having separate workflows can encounter a case where the publish to `docs.nvidia.com` fails, but redirects are deployed which will be broken. 
- Updated all references of `/deployment/stable` to `/deployment/latest` following the new versioning scheme.

josephine-wolf-oberholtzer (approved):

PR #833, opened 2026-09-02
Redirect migrated API docs to docs.nvidia.com

bdice:
## Summary

Redirect API roots directly to each migrated library's `docs.nvidia.com/<library>/` site. Redirect stable and nightly paths to `docs.nvidia.com/<library>/latest/`, preserving nested paths. Stop importing externally hosted versions from S3 while retaining pre-migration docs.

Retire the user-facing legacy, stable, and nightly documentation labels. Label the newest documentation `Latest`, label other links by release number, and update the API page definitions accordingly.

This is the full migration rollout. Use the Dask-CUDA-only draft in #834 to test the redirect approach first.

### DIFF
--- a/extensions/rapids_docs/api.py
+++ b/extensions/rapids_docs/api.py
@@ -31,9 +31,13 @@ def _api_docs(data: dict, section: str) -> str:
         if not external_docs_url:
             for name in ("nightly", "stable", "legacy"):
                 if project["versions"].get(name) == 1:
-                    label = _version_label(project, name, data["releases"])
-                    url = _documentation_url(project, name, label)
-                    versions.append(f"[{name.title()} ({label})]({url})")
+                    version = _version_label(project, name, data["releases"])
+                    url = _documentation_url(project, name, version)
+                    if project["first_docs_nvidia_com_release"]:
+                        label = "Latest" if name == "nightly" else version
+                    else:
+                        label = f"{name.title()} ({version})"
+                    versions.append(f"[{label}]({url})")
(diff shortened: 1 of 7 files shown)


extensions/rapids_docs/api.py
```diff
@@ -31,9 +32,10 @@ def _api_docs(data: dict, section: str) -> str:
         if not external_docs_url:
             for name in ("nightly", "stable", "legacy"):
                 if project["versions"].get(name) == 1:
-                    label = _version_label(project, name, data["releases"])
-                    url = _documentation_url(project, name, label)
-                    versions.append(f"[{name.title()} ({label})]({url})")
+                    version = _version_label(project, name, data["releases"])
+                    url = _documentation_url(project, name, version)
+                    label = "Latest" if name == "nightly" else version
```
josephine-wolf-oberholtzer, line 37: Confirming uppercase `Latest` and not `latest`?

josephine-wolf-oberholtzer (approved):
LGTM, just confirming you wanna see `Latest` in the label and not `latest`.

extensions/rapids_docs/api.py
bdice: I’m case-insensitive on this. What do you think? I like Latest here and latest in the version picker but I can’t justify why. Just feels.

josephine-wolf-oberholtzer: Feels is fine. Can always cut a second PR if you change your mind.

PR #834, opened 2026-09-02
Redirect Dask-CUDA docs to docs.nvidia.com

bdice:
## Summary

Redirect the Dask-CUDA API root directly to `docs.nvidia.com/dask-cuda/`. Redirect stable and nightly paths to `docs.nvidia.com/dask-cuda/latest/`, preserving nested paths. Stop importing Dask-CUDA stable and nightly docs from S3 while retaining legacy docs.

This is a limited rollout for testing before the full migration in #833.


josephine-wolf-oberholtzer (approved):
Let's try it.

PR #835, opened 2026-09-09
Point C++ API docs to NVIDIA Sphinx sites

bdice:
## Summary

Route the libcudf, libkvikio, and librapidsmpf API cards from old Doxygen sites to the new Sphinx C++ documentation on docs.nvidia.com beginning with release 26.10. Older releases continue to use their existing docs.rapids.ai links, and migrated builds are no longer imported into the portal.


ncclementi (approved):
LGTM

PR #836, opened 2026-09-09
add redirects and remove S3 download for deployment docs

jayavenkatesh19:
Adds redirects and removes the S3 download of the deployment docs from the old `rapids-docs` S3 bucket. Deployment docs are now served directly on `docs.nvidia.com` from the new bucket

Redirects are mirrored from #833


_redirects
```diff
@@ -95,8 +95,12 @@
...
 /api/ucxx /api/ucxx/stable/
 /api/ucxx/ /api/ucxx/stable/
-/deployment /deployment/stable/
-/deployment/ /deployment/stable/
+/deployment https://docs.nvidia.com/datascience/deployment/latest/ 301!
+/deployment/ https://docs.nvidia.com/datascience/deployment/latest/ 301!
+/deployment/nightly https://docs.nvidia.com/datascience/deployment/latest/ 301!
+/deployment/nightly/* https://docs.nvidia.com/datascience/deployment/latest/:splat 301!
+/deployment/stable https://docs.nvidia.com/datascience/deployment/latest/ 301!
+/deployment/stable/* https://docs.nvidia.com/datascience/deployment/latest/:splat 301!
```
ncclementi, line 103: Don't quite sure how things work, but checking just in case. 
Do we need all these cases, or can we just modify the one above to maybe be something like? 

```
/deployment/* https://docs.nvidia.com/datascience/deployment/latest/:splat 301!
```

jayavenkatesh19: Oh that would not work, because `:splat` is plain wildcard matching. 

So `/deployment/stable/cloud/aws/eks/` would turn into `docs.nvidia.com/latest/stable/cloud/aws/eks` which is a non existing page. 

If the versioning scheme remained the same, one rule would have solved this but as we are changing that we need this. I got these changes off the other redirects too, and all the API docs have the same rules as well.

ncclementi (approved):
LGTM

PR #838, opened 2026-09-16
Document historical XGBoost conda channel changes in RSN 62

bdice:
## Summary

- Document the move of historical XGBoost packages to `rapidsai/label/legacy-xgboost` so strict channel priority permits newer conda-forge releases.
- Explain patched `rapids` metapackages for v25.12–v26.08 and the loss of XGBoost RMM integration by default.
- Provide migration and legacy restoration guidance, including the actual install-time warning.


notices/rsn0062.md
```diff
@@ -19,24 +19,91 @@ notice_status_color: yellow
...
+It can be restored by installing:
+  conda install -c rapidsai/label/legacy-xgboost "rapids=26.08" "xgboost=*=rapidsai_*"
+See https://docs.nvidia.com/datascience/notices/rsn0062/ for more information.
+```
+
+The warning omits `-c rapidsai`; include it when running that command to resolve
+the explicitly requested `rapids` metapackage.
+
+The original, unpatched build-0 `rapids` metapackages for these releases will be
+moved to `rapidsai/label/broken`. Installation specifications pinned to those builds
```
msarahan:
```suggestion
moved to `rapidsai/label/legacy-xgboost`. Installation specifications pinned to those builds
```

msarahan (approved):

notices/rsn0062.md
```diff
@@ -19,24 +19,91 @@ notice_status_color: yellow
...
-wheels from the Python Package Index `PyPI`.
+### Use standard XGBoost packages
+
+Existing users should migrate workloads away from `rapids-xgboost`. Use the
+`xgboost` package from `conda-forge`, or the wheels from the Python Package Index
+`PyPI`. These packages do not include RAPIDS RMM integration.
+
+In conda installation commands and environment files, replace explicit
+`rapids-xgboost` or `rapidsai::xgboost` requirements with `conda-forge::xgboost`,
+using a version compatible with your workload.
```
jakirkham, line 90: Think we will want to expand a bit on some of the notable differences between the RAPIDS XGBoost package and the conda-forge one. Here are some initial thoughts on how we might do this. Open to revisions that capture this general gist


> In conda installation commands and environment files, replace explicit
> `rapids-xgboost` or `rapidsai::xgboost` requirements with `conda-forge::xgboost`,
> using a version compatible with your workload.
> 
> There are [separate CPU & CUDA builds of XGBoost in conda-forge]( https://github.com/conda-forge/xgboost-feedstock/blob/14ad7ccb4c4424919525a1d1e9fe9c258f491eae/recipe/meta.yaml#L6 ) (as opposed to the CUDA-only builds in RAPIDS). So users will need to specify CUDA to get a GPU enabled package.
> 
> As usual please make sure to [specify the CUDA Toolkit version desired with the `cuda-version` package]( https://github.com/conda-forge/cuda-version-feedstock/blob/main/recipe/README.md#for-users-of-gpu-packages ) when creating the environment. Then install the XGBoost package in that environment.
> 
> conda install -c conda-forge 'xgboost=*=cuda*'
> 
> Note: The conda-forge XGBoost requires that the CUDA driver be installed and be detectable by Conda. Users can check the CUDA driver version Conda detects by running `conda info` and looking at the `virtual packages` section for `__cuda`. If `__cuda` is not present there or defined as `__cuda=0=0`, Conda is not able to detect the driver version (likely because the CUDA driver is not installed). Users can always override the value of `__cuda` by setting the environment variable `CONDA_OVERRIDE_CUDA` to the `<MAJOR>.<MINOR>` version of the latest supported CUDA Toolkit by the driver.

Side note: The suggestion feature fell apart with this text. So gave up and stuck it in a quote

jakirkham (changes requested):
Thanks Bradley! 🙏

These look like good improvements that we should add.

Think we need to call out some important differences between the RAPIDS XGBoost and conda-forge XGBoost packages. Like the hard requirement on the CUDA driver being present in conda-forge. Have proposed some wording below. Open to revising it as long as we capture the main points.

notices/rsn0062.md
bdice: I'd appreciate if you can push directly here, if you don't mind. Thanks!

bdice: Done in d3d31ad.

